PR #3, opened 2019-10-11
Add edit view and patching for IS-05 resources

rufusu-sony:
This PR introduces an edit view for receivers and senders.

- Manual connections can be made by inputting transport parameters or using transport files.
- Tabs are now material-ui components
- Connection API endpoints are now retrieved in parallel
- Only changes made to the API response are patched


garethsb-sony:
This is looking really nice.

Don't think you intended the themeing stuff to be in this PR?

And there's quite a bit of console noise, e.g.

> index.js:2178 Warning: Each child in a list should have a unique "key" prop.
> Check the render method of `RTPReceiver`. See https://fb.me/react-warning-keys for more information.
(and the same for `CardFormIterator`)

> index.js:2178 Warning: Failed prop type: Invalid prop `spacing` of value `3` supplied to `Grid`, expected one of [0,8,16,24,32,40].

garethsb-sony:
I've fixed most of the warnings.

We can redo the Card Grid spacing if we end up with latest Material UI.

I think using the Material UI Icons for the SelectInput/SelectField seems to be resulting in:
> index.js:2178 Warning: Missing translation for key: "[object Object]"

garethsb-sony:
Really significant new feature! Thanks, Rufus!

PR #4, opened 2019-10-15
Add dark theme and use material ui inputs

rufusu-sony:
- The settings page has been re-written to have a theme consistent with
  the rest of the app using material ui inputs.
- A dark theme has been added and theming inconsistencies removed.
- HTTP errors returned when the query API is not found will no longer
  crash the entire app.
- The scrollbar is always shown on PC to prevent jarring motion.

### DIFF
--- a/Development/src/components/JSONViewer.js
+++ b/Development/src/components/JSONViewer.js
@@ -11,7 +11,7 @@ import get from 'lodash/get';
 import ReactJson from 'react-json-view';
 
 // Base16 dark theme color palette
-const rjv_dark = {
+const rjvDark = {
     base00: 'rgba(0, 0, 0, 0)',
     base01: '#282828',
     base02: '#383838',
@@ -36,6 +36,11 @@ export default function JSONViewer({ endpoint, ...controllerProps }) {
         setChecked(prev => !prev);
     };
     const theme = useTheme();
+    let rjvTheme;
+    if (theme.palette.type === 'dark') {
+        rjvTheme = rjvDark;
+    } else rjvTheme = 'rjv-default';
+
     return (
         <div>
             <FormControlLabel
@@ -45,28 +50,10 @@ export default function JSONViewer({ endpoint, ...controllerProps }) {
             <Collapse in={checked}>
                 <Card>
                     <CardContent>
-                        {(() => {
-                            if (theme.palette.type === 'dark') {
-                                return (
-                                    <ReactJson
-                                        src={get(
-                                            controllerProps.record,
-                                            endpoint
-                                        )}
-                                        theme={rjv_dark}
-                                    />
-                                );
-                            } else {
-                                return (
-                                    <ReactJson
-                                        src={get(
-                                            controllerProps.record,
-                                            `${endpoint}`
-                                        )}
-                                    />
-                                );
-                            }
-                        })()}
+                        <ReactJson
+                            src={get(controllerProps.record, endpoint)}
+                            theme={rjvTheme}
+                        />
                     </CardContent>
                 </Card>
             </Collapse>
--- a/Development/src/components/JSONViewer.js
+++ b/Development/src/components/JSONViewer.js
@@ -1,3 +1,4 @@
+import React from 'react';
 import {
     Card,
     CardContent,
@@ -5,15 +6,36 @@ import {
     FormControlLabel,
     Switch,
 } from '@material-ui/core';
+import { useTheme } from '@material-ui/styles';
 import get from 'lodash/get';
-import React from 'react';
 import ReactJson from 'react-json-view';
+
+// Base16 dark theme color palette
+const rjv_dark = {
+    base00: 'rgba(0, 0, 0, 0)',
+    base01: '#282828',
+    base02: '#383838',
+    base03: '#585858',
+    base04: '#b8b8b8',
+    base05: '#d8d8d8',
+    base06: '#e8e8e8',
+    base07: '#f8f8f8',
+    base08: '#be4b3c',
+    base09: '#dc9656',
+    base0A: '#f7ca88',
+    base0B: '#a1b56c',
+    base0C: '#86c1b9',
+    base0D: '#a16946',
+    base0E: '#aa759f',
+    base0F: '#7cafc2',
+};
 
 export default function JSONViewer({ endpoint, ...controllerProps }) {
     const [checked, setChecked] = React.useState(false);
     const handleChange = () => {
         setChecked(prev => !prev);
     };
+    const theme = useTheme();
     return (
         <div>
             <FormControlLabel
@@ -23,9 +45,28 @@ export default function JSONViewer({ endpoint, ...controllerProps }) {
             <Collapse in={checked}>
                 <Card>
                     <CardContent>
-                        <ReactJson
-                            src={get(controllerProps.record, endpoint)}
-                        />
+                        {(() => {
+                            if (theme.palette.type === 'dark') {
+                                return (
+                                    <ReactJson
+                                        src={get(
+                                            controllerProps.record,
+                                            endpoint
+                                        )}
+                                        theme={rjv_dark}
+                                    />
+                                );
+                            } else {
+                                return (
+                                    <ReactJson
+                                        src={get(
+                                            controllerProps.record,
+                                            `${endpoint}`
+                                        )}
+                                    />
+                                );
+                            }
+                        })()}
                     </CardContent>
                 </Card>
             </Collapse>
--- a/Development/src/components/URLField.js
+++ b/Development/src/components/URLField.js
@@ -1,0 +1,21 @@
+import React from 'react';
+import { Typography } from '@material-ui/core';
+import get from 'lodash/get';
+
+const UrlField = ({ source, record = {} }) => (
+    <a
+        href={get(record, source)}
+        target="_blank"
+        rel="noopener noreferrer"
+        style={{ textDecoration: 'none' }}
+    >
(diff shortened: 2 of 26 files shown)


garethsb-sony:
Seeing a new dev console log message:
> <Provider> does not support changing `store` on the fly. It is most likely that you see this error because you updated to Redux 2.x and React Redux 2.x which no longer hot reload reducers automatically. See https://github.com/reactjs/react-redux/releases/tag/v2.0.0 for the migration instructions.

garethsb-sony:
I don't know why but I expected the Dark Theme toggle to change the theme immediately.... obviously there's a Save button though... Hmm...

garethsb-sony:
RQL toggle and Dark Theme toggle are so unrelated I suggest we move them apart.
E.g. a `<HorizontalRule />` after the RQL toggle, maybe?

garethsb-sony:
Another new dev console message:

> Warning: validateDOMNesting(...): <p> cannot appear as a descendant of <p>.
    in p (created by Typography)
    in Typography (created by WithStyles(Typography))
    in WithStyles(Typography) (at TAIField.js:10)
    in p (created by Typography)
    in Typography (created by WithStyles(Typography))
    in WithStyles(Typography) (at TAIField.js:7)
    in TAIField (at nodes.js:293)

Looks like, in adopting the better, themed approach, the `TAIField` has in any case lost some of its formatting (date-time is appearing on a second line and unparenthesized, both of which seem to make it a little harder to read).

garethsb-sony:
This one may not be new, but the little data-grids for e.g. API Endpoints in the Node show page probably need sorting turning off, because we don't support that. At the moment, clicking the column headings results in:

> Uncaught TypeError: _this.props.setSort is not a function
    at Datagrid._this.updateSort

(Minor: also noticed 'Ref type' in the 'Clocks' table ought to be 'Ref Type' for consistency with other titles' capitalization.)

garethsb-sony:
On Receiver show page > Summary, in dark theme, 'Interface Bindings' and 'Caps Media Types' are black text rather than white text like other fields.

garethsb-sony:
> (Minor: also noticed 'Ref type' in the 'Clocks' table ought to be 'Ref Type' for consistency with other titles' capitalization.)

Similarly (and again nothing to do with this PR), on the Settings > Query APIs show page, several of the fields need label overrides.

garethsb-sony:
On my screen, the cards within the Show and Edit views for Senders and Receivers seem to have lost some of their definition (top edge seems almost imperceptible). I think that's new?

Development/src/components/JSONViewer.js
```diff
@@ -1,19 +1,41 @@
...
     FormControlLabel,
     Switch,
 } from '@material-ui/core';
+import { useTheme } from '@material-ui/styles';
 import get from 'lodash/get';
-import React from 'react';
 import ReactJson from 'react-json-view';
 
+// Base16 dark theme color palette
+const rjv_dark = {
```
garethsb-sony: What does 'rjv' stand for? Where do these values come from?

Development/src/components/JSONViewer.js
```diff
@@ -23,9 +45,28 @@ export default function JSONViewer({ endpoint, ...controllerProps }) {
...
+                                        )}
+                                        theme={rjv_dark}
+                                    />
+                                );
+                            } else {
+                                return (
+                                    <ReactJson
+                                        src={get(
+                                            controllerProps.record,
+                                            `${endpoint}`
```
garethsb-sony: Just `endpoint` as above?

garethsb-sony:
Edit button still has the text underlined (again, not down to this PR, I know)

Development/src/pages/Dashboard.js
```diff
@@ -10,8 +10,8 @@ import Settings from '../settings';
 export default () => (
     <div>
         <Card>
-            <Typography variant="display3" component="h2" align="center">
-                Welcome to the nmos-js Client
+            <Typography variant="display3" align="center">
+                Welcome to the nmos&#8209;js Client
```
garethsb-sony: :+1: Shame there isn't an HTML entity like `&nbsp;` for non-breaking hyphen!

Development/src/components/URLField.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { Typography } from '@material-ui/core';
+import get from 'lodash/get';
+
+const UrlField = ({ source, record = {} }) => (
```
garethsb-sony: UrlField or URLField? Should the filename and component name match?
And be consistent with `TAIField` (or `TaiField` if we're going that way...).

Development/src/pages/nodes.js
```diff
@@ -228,7 +229,7 @@ const ItemArrayField = ({ className, source, record = {} }) => (
     <div>
         {get(record, source).map(item => (
             <div key={item} className={className}>
-                {item}
+                <Typography>{item}</Typography>
```
garethsb-sony: Is the outer `div` still required or could the `key` and `className` be added to the `Typography`?

Development/src/theme/ThemeContext.js
```diff
@@ -0,0 +1,41 @@
...
+export const AppThemeProvider = ({ children }) => {
+    const [themeState, setThemeState] = React.useState({
+        mode: cookies.get('theme'),
+    });
+    if (themeState.mode === undefined) setThemeState({ mode: 'light' });
+
+    const theme = createMuiTheme({
+        palette: {
+            primary: lightBlue,
+            secondary: lightBlue,
```
garethsb-sony: Probably seems a bit wrong to use the same colour for primary and secondary?

rufusu-sony: The secondary colour is used in the same way as the primary in dark mode. I wanted to keep elements light blue, however we could use a different colour when in dark mode.

Development/src/components/URLField.js
rufusu-sony: The name UrlField was used as the component was a drop in replacement for the react-admin version, exception being it themes correctly. I will change to URLField to be more consistent.

Development/src/pages/Dashboard.js
rufusu-sony: I'll leave a comment on the line above

Development/src/components/JSONViewer.js
rufusu-sony: rjv stands for react-json-viewer. Base 16 is a theming standard for code https://github.com/chriskempson/base16 
I'll leave a comment and maybe change the name

Development/src/theme/ThemeContext.js
garethsb-sony: Let's discuss, one more shade in the light mode might not be a bad thing

Development/src/components/URLField.js
garethsb-sony: Let's discuss. Not sure whether reusing names is frowned upon in React? If not, maybe we should be going the other way and making it TaiField...

garethsb-sony:
Another dev console message I hadn't seen before:

> Warning: Received NaN for the `value` attribute. If this is expected, cast the value to a string.

Seems to be related to the 'Paging Limit' which in my browser is empty.

garethsb-sony:
> This one may not be new, but the little data-grids for e.g. API Endpoints in the Node show page probably need sorting turning off, because we don't support that. At the moment, clicking the column headings results in:
> 
> > Uncaught TypeError: _this.props.setSort is not a function
> > at Datagrid._this.updateSort

This would seem just like one of the comments in https://github.com/marmelab/react-admin/issues/3157 but that's allegedly fixed in react-admin 2.9.0. Hmmm.

garethsb-sony:
Using `Divider` rather than my quick'n'dirty `HorizontalRule` has made this warning come back:

> Warning: React does not recognize the `basePath` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `basepath` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
    in hr (created by Divider)
    in Divider (created by WithStyles(Divider))
    in WithStyles(Divider) (at SenderTransportParams.js:538)

garethsb-sony:
> On my screen, the cards within the Show and Edit views for Senders and Receivers seem to have lost some of their definition (top edge seems almost imperceptible). I think that's new?

I think this is just a problem with layering a card on top of a card, so there's no difference in the background colour. Let's not worry right now!

garethsb-sony:
> On Receiver show page > Summary, in dark theme, 'Interface Bindings', 'Caps Media Types' and 'Caps Event Types' are black text rather than white text like other fields.

Fixed.

> (Minor: also noticed 'Ref type' in the 'Clocks' table ought to be 'Ref Type' for consistency with other titles' capitalization.)

Fixed.

> Similarly (and again nothing to do with this PR), on the Settings > Query APIs show page, several of the fields need label overrides.

Fixed.

> Another new dev console message:
>
> > Warning: validateDOMNesting(...): <p> cannot appear as a descendant of <p>.

Fixed.

> Looks like, in adopting the better, themed approach, the TAIField has in any case lost some of its formatting (date-time is appearing on a second line and unparenthesized, both of which seem to make it a little harder to read).

Fixed.

> I don't know why but I expected the Dark Theme toggle to change the theme immediately.... 
> RQL toggle and Dark Theme toggle are so unrelated I suggest we move them apart.

Both fixed by the move to the AppBar. :-)

garethsb-sony:
> Edit button still has the text underlined (again, not down to this PR, I know)

Think I've fixed that using `component`.

garethsb-sony:
Thanks, Rufus, we have dark theme, and much nicer 'Settings'!!

PR #5, opened 2019-10-18
Fetch from multiple connection API URLs 

rufusu-sony:
- Attempt to fetch from each connection API URL until successful. This
  fits the case where the node is not accessible on some network
  interfaces.
- Error handling in the dataProvider will no longer crash the app.
- The Active, Staged and Transportfile tabs are now shown conditionally.
  If no controls are found the tabs are not shown, if unable to connect
  to controls, the tabs are disabled. This change also applies to the
  edit button.
- A warning has been fixed for nesting a `<pre>` inside `<Typography>`

### DIFF
--- a/Development/src/dataProvider.js
+++ b/Development/src/dataProvider.js
@@ -401,7 +401,7 @@ async function convertHTTPResponseToDataProvider(
                 let connectionAddresses = {};
                 deviceJSONData.controls.forEach(function(control) {
                     const type = control.type.replace('.', '_');
-                    if (type.includes('sr-ctrl')) {
+                    if (type.startsWith('urn:x-nmos:control:sr-ctrl')) {
                         if (!connectionAddresses.hasOwnProperty(type)) {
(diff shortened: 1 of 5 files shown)


Development/src/dataProvider.js
```diff
@@ -371,28 +399,50 @@ async function convertHTTPResponseToDataProvider(
...
+                            set(connectionAddresses, type, [control.href]);
+                        } else {
+                            connectionAddresses[type].push(control.href);
+                        }
+                    }
+                });
+                // Return IS-04 if no Connection API endpoints
+                if (
+                    Object.keys(connectionAddresses).length === 0 &&
+                    connectionAddresses.constructor === Object
```
garethsb-sony: Minor: I guess you copied this condition from somewhere like https://stackoverflow.com/questions/679915/how-do-i-test-for-an-empty-javascript-object, but the rationale given for adding the `.constructor` check isn't applicable here.

PR #8, opened 2019-11-05
Code quality & CI

rufusu-sony:
- List views now use React's useState hook.
- Pagination on the log view now uses the user's preference
- Settings and Dashboard have been combined
- Update dependencies
- Add TravisCI
- Add Dockerfile
- Remove unused files


.travis.yml
```diff
@@ -0,0 +1,10 @@
+language: node_js
+node_js: lts/*
+
+before_install: cd Development
+
+install: yarn install
+
+script:
+  - yarn run build
+  - yarn test --passWithNoTests --coverage --watchAll=false
```
garethsb-sony: Failing builds if linting reported error would be good!

Development/src/pages/logs.js
```diff
@@ -37,178 +37,156 @@ const cardActionStyle = {
...
-    }
-
-    async firstLoad() {
+export const EventsList = () => {
+    const firstLoad = async () => {
         const params = {
             filter: {},
             pagination: { page: 1, perPage: 10 },
-            sort: { field: 'timestamp', order: 'DESC' },
+            sort: { field: 'id', order: 'DESC' },
```
garethsb-sony: I don't think `pagination` or `sort` should be here at all, since they're not relevant for cursor-based pagination with our `dataProvider`?

Development/src/pages/receivers.js
```diff
@@ -321,14 +303,18 @@ const ShowSummaryTab = ({ controllerProps, ...props }) => {
                     />
                 )}
                 {controllerProps.record &&
-                    controllerProps.record.caps.media_types && (
+                    controllerProps.record.hasOwnProperty(
+                        'record.caps.media_types'
```
garethsb-sony: Check this!

Development/src/pages/subscriptions.js
```diff
@@ -31,164 +31,146 @@ import JsonIcon from '../components/JsonIcon';
...
-        this.firstLoad();
-    }
-
-    async firstLoad() {
+export const SubscriptionsList = () => {
+    const firstLoad = async () => {
         const params = {
             filter: {},
             pagination: { page: 1, perPage: 10 },
-            sort: { field: 'timestamp', order: 'DESC' },
```
garethsb-sony: Don't think `pagination` should be here either as it is only relevant to offset-limit pagination, and the NMOS APIs don't support sorting either, so no `sort` either.

garethsb-sony:
Also, please add the simple Travis CI badge to the README, as previously done for sony/nmos-cpp.

Development/src/pages/logs.js
rufusu-sony: These params are actually present on all list views, as it has no effect on our dataProvider, I will remove the extraneous lines.

PR #13, opened 2019-11-20
Add connect tab to receiver show view

rufusu-sony:
- A new tab has been added in the receiver show view with a list of
  compatible senders based on transport type.
- A connect button auto populates and sends an activation request to the
  receiver staged endpoint with the transport params of the chosen
  sender.
- Show the subscription_active field on both receiver and sender list
  views for query api versions >= v1.2

### DIFF
--- a/Development/src/components/MakeConnection.js
+++ b/Development/src/components/MakeConnection.js
@@ -1,0 +1,217 @@
+import setJSON from 'json-ptr';
+import cloneDeep from 'lodash/cloneDeep';
+import get from 'lodash/get';
+import set from 'lodash/set';
+
+const oneToOneTransportParams = {
+    'urn:x-nmos:transport:mqtt': [
+        'broker_protocol',
+        'broker_authorization',
+        'broker_topic',
+        'connection_status_broker_topic',
+    ],
+    'urn:x-nmos:transport:rtp': [
+        'rtp_enabled',
+        'source_ip',
+        'destination_port',
+        'fec_enabled',
+        'fec_destination_ip',
+        'fec_mode',
+        'fec1D_destination_port',
+        'fec2D_destination_port',
+        'rtcp_enabled',
+        'rtcp_destination_ip',
+        'rtcp_destination_port',
+    ],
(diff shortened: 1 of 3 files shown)


Development/src/pages/receivers.js
```diff
@@ -233,6 +250,12 @@ export const ReceiversShow = props => {
                                         }
                                     />
                                 ))}
+                            <Tab
+                                label="Connect"
+                                value={`${props.match.url}/connect`}
+                                component={Link}
+                                to={`${props.basePath}/${props.id}/show/connect`}
+                            />
```
garethsb-sony: Should this tab be `disabled` if the Connection API was inaccessible?

Development/src/components/MakeConnection.js
```diff
@@ -0,0 +1,217 @@
...
+                return data;
+            })
+            .then(data => {
+                if (get(data, 'sender') === undefined) reject();
+                if (get(data, 'receiver') === undefined) reject();
+                if (
+                    get(data.sender, '$transporttype') !==
+                    get(data.receiver, '$transporttype')
+                )
+                    reject();
```
garethsb-sony: Braces, please

Development/src/components/MakeConnection.js
```diff
@@ -0,0 +1,217 @@
...
+                    lhs,
+                    true
+                );
+            }
+        });
+    });
+};
+
+export const isMulticast = address => {
+    if (
```
garethsb-sony: `return a || b;` maybe?

Development/src/pages/receivers.js
```diff
@@ -544,3 +578,199 @@ const EditStagedTab = props => (
...
+
+    const nextPage = async label => {
+        const data = await dataProvider(label, 'senders');
+        setSendersListData(data);
+    };
+
+    // receiverData initialises undefined, update when no longer null
+    useEffect(() => {
+        if (receiverData !== null)
+            setParams({
```
garethsb-sony: Does this _combine_ with any user-specified filter as we want, or would they replace each other?

PR #15, opened 2019-11-27
react-admin major version bump & list view redux integration

rufusu-sony:
- Edit, List and View components have been refactored into separate files.
- Imports from ra-ui-materialui and ra-core have been changed to react-admin to reduce the number of required dependencies in package.json.
- ChipConditionalLabel & QueryVersion have been refactored into a reusable components.
- react-admin has been upgraded to version 3.0.0 and all breaking changes have been resolved.
- Added useGetList to hook into the react-admin redux store. This allows the refresh button to trigger a refresh and other parts of the app to retreive the data through a store.
- Pagination is now acheived using a param instead of a type prop for compatibility with the react admin redux actions.


Development/src/components/MakeConnection.js
```diff
@@ -2,6 +2,7 @@ import setJSON from 'json-ptr';
 import cloneDeep from 'lodash/cloneDeep';
 import get from 'lodash/get';
 import set from 'lodash/set';
+import dataProvider from '../dataProvider';
```
rufusu-sony: As MakeConnection is not a React functional component, hooks cannot be used. A dataProvider prop could be passed from a higher level component however?

Development/src/components/useGetList.js
```diff
@@ -0,0 +1,102 @@
...
+        state =>
+            state.admin.resources[resource]
+                ? state.admin.resources[resource].list.total
+                : null
+    );
+    const data = useSelector(
+        state =>
+            state.admin.resources[resource]
+                ? state.admin.resources[resource].data
+                : null,
```
rufusu-sony: Unsure on the usefulness of returning null as this will crash the app if ids.map is reached

Development/src/pages/devices/DevicesShow.js
```diff
@@ -0,0 +1,133 @@
...
+import JsonIcon from '../../components/JsonIcon';
+import MapTags from '../../components/TagsField';
+import TAIField from '../../components/TAIField';
+import UrlField from '../../components/URLField';
+import QueryVersion from '../../components/QueryVersion';
+import ChipConditionalLabel from '../../components/ChipConditionalLabel';
+
+const cookies = new Cookies();
+
+const DevicesTitle = ({ record }) => {
```
rufusu-sony: May be worth refactoring ResourceTitle into it's own component

Development/src/pages/devices/DevicesShow.js
```diff
@@ -0,0 +1,133 @@
...
+            {record
+                ? record.label
+                    ? `${record.label}`
+                    : `${record.id}`
+                : 'Unknown'}
+        </span>
+    );
+};
+
+const DevicesShowActions = ({ basePath, data, resource }) => (
```
rufusu-sony: Same as ResourceTitle, needs a refactor

Development/src/pages/logs.js
```diff
@@ -200,7 +183,7 @@ const EventsShowActions = ({ basePath, data, resource }) => (
             </Button>
         ) : null}
         <ListButton title={'Return to ' + basePath} basePath={basePath} />
-    </CardActions>
+    </TopToolbar>
 );
 
 export const EventsShow = props => (
```
rufusu-sony: May be worth refactoring logs into list and show view like other resources

Development/src/pages/menu.js
```diff
@@ -24,118 +23,103 @@ import SettingsInputAntennaIcon from '@material-ui/icons/SettingsInputAntenna';
...
+                <ListItemIcon>
+                    <AnnouncementIcon />
+                </ListItemIcon>
+                <ListItemText primary="Logs" />
+            </StyledListItem>
+        </MenuList>
     );
 };
 
-NestedList.propTypes = {
```
rufusu-sony: This has been removed as it was an outlier however it would be beneficial to check prop types on all components

Development/src/theme/ThemeContext.js
```diff
@@ -17,13 +17,18 @@ export const AppThemeProvider = ({ children }) => {
     });
     if (themeState.mode === undefined) setThemeState({ mode: 'light' });
 
-    const theme = createMuiTheme({
+    let theme = createMuiTheme({
```
rufusu-sony: Theme declaration can be condensed to:
`const theme = responsiveFontSizes(createMuiTheme(`

Development/src/pages/logs.js
garethsb-sony: Yes, consistency would be nice.

Development/src/App.js
```diff
@@ -3,27 +3,27 @@ import { Admin, Resource } from 'react-admin';
 import { useTheme } from '@material-ui/styles';
 
 import Dashboard from './pages/settings';
+import { ReceiversEdit, ReceiversList, ReceiversShow } from './pages/receivers';
```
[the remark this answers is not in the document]
garethsb-sony: Yeah, I spotted another one of these in SenderTransportParams. Alphabetizing might be a good idea if the linter supports checking that?

Development/src/pages/queryapis/QueryAPIsList.js
```diff
@@ -0,0 +1,32 @@
+import React, { Fragment } from 'react';
+import { Datagrid, List, ShowButton, TextField } from 'react-admin';
+import ConnectButton from '../../components/ConnectButton';
+
+const QueryAPIsTitle = ({ record }) => {
+    return <span>Query APIs{record ? `: ${record.name}` : ''}</span>;
```
garethsb-sony: Just noting that this is different in style to the equivalent code in the NMOS Resources, so if we're to add `ResourceTitle` as a component, let's work out why (no good reason, I'm guessing!)

Development/src/pages/devices/DevicesList.js
```diff
@@ -0,0 +1,113 @@
...
+        setPaginationCursor(null);
+        setSeed(Math.random());
+    };
+
+    if (!loaded) return <Loading />;
+    if (error) return <Error />;
+    if (!data) return null;
+
+    return (
+        <>
```
garethsb-sony: Is this weird `<>`...`</>` construct, the fragment syntax?

Development/src/pages/logs.js
```diff
@@ -150,7 +139,7 @@ export const EventsList = () => {
                             </TableRow>
                         </TableHead>
                         <TableBody>
-                            {data.data.map(item => (
+                            {data.map(item => (
```
garethsb-sony: Well that looks nicer, LOL!

Development/src/pages/logs.js
```diff
@@ -1,92 +1,81 @@
...
+            delete currentFilter[name];
         }
-        const filteredDataObject = await dataProvider('GET_LIST', 'events', {
-            filter: filter,
-        });
-        setFilterState(filter);
-        setData(filteredDataObject);
+        setFilter(currentFilter);
+        setPaginationCursor(null);
+        setSeed(Math.random());
```
garethsb-sony: Let's discuss short-term explanation/long-term refactor for this stuff.

Development/src/pages/logs.js
```diff
@@ -174,20 +163,14 @@ export const EventsList = () => {
                     <PaginationButton label="PREV" nextPage={nextPage} />
                     <PaginationButton label="NEXT" nextPage={nextPage} />
                     <PaginationButton label="LAST" nextPage={nextPage} />
-                    <Button
```
garethsb-sony: So you decided that with the Refresh button working again, we might as well get rid of Clear All Filters everywhere?

Development/src/pages/senders/SendersEdit.js
```diff
@@ -0,0 +1,158 @@
...
+);
+
+const SendersEdit = props => {
+    const theme = useTheme();
+    const tabBackgroundColor =
+        theme.palette.type === 'light'
+            ? theme.palette.grey[100]
+            : theme.palette.grey[900];
+    return (
+        <Fragment>
```
garethsb-sony: OK, no, I don't get it, why e.g. `<Fragment>` here, but `<>` in SendersShow?

Development/src/pages/receivers/ReceiversShow.js
```diff
@@ -0,0 +1,538 @@
...
+
+    // receiverData initialises undefined, update when no longer null
+    useEffect(() => {
+        if (receiverData !== null)
+            setFilter({ transport: get(receiverData, 'transport') });
+    }, [receiverData]);
+
+    const connect = (senderID, receiverID, endpoint) => {
+        MakeConnection(senderID, receiverID, endpoint, props).then(() => {
+            refresh();
```
garethsb-sony: New call to `refresh()` here?

Development/src/pages/nodes/NodesShow.js
```diff
@@ -0,0 +1,159 @@
...
+                <UrlField source="href" label="Address" />
+                <TextField source="hostname" />
+                {controllerProps.record && QueryVersion() >= 'v1.1' && (
+                    <ItemArrayField
+                        label="API Versions"
+                        source="api.versions"
+                    />
+                )}
+                {controllerProps.record && QueryVersion() >= 'v1.1' && (
+                    <ArrayField label=" API Endpoints" source="api.endpoints">
```
garethsb-sony: Why _is_ there an extra space at the start of the label? (I see it was there already!)

Development/src/pages/nodes/NodesShow.js
```diff
@@ -0,0 +1,159 @@
...
+                        <TextField source="type" />
+                    </Datagrid>
+                </ArrayField>
+                {controllerProps.record && QueryVersion() >= 'v1.2' && (
+                    <ArrayField source="interfaces">
+                        <Datagrid>
+                            <TextField source="chassis_id" label="Chassis ID" />
+                            <TextField source="name" />
+                            <TextField source="port_id" label="Port ID" />
+                        </Datagrid>
```
garethsb-sony: (Still need the separate PR to add `attached_network_device.chassis_id` and `.port_id` if `>= 'v1.3'`. Think we decided the full name was too long so we could show "Attached Chassis ID" and "Attached Port ID" - or maybe "Remote Chassis ID" and "Remote Port ID" since LLDP uses the terms local and remote.)

Development/src/pages/logs.js
```diff
@@ -1,92 +1,81 @@
 import React, { useState } from 'react';
-import {
```
garethsb-sony: This file needs splitting out into LogsShow.js and LogsList.js like the others, right?

Development/src/LayoutComponent.js
```diff
@@ -3,8 +3,8 @@ import { Layout } from 'react-admin';
 import NestedList from './pages/menu';
```
garethsb-sony: I'm so happy you renamed this. It's bugged me since day one. :laughing:

Development/src/App.js
```diff
@@ -3,27 +3,27 @@ import { Admin, Resource } from 'react-admin';
...
-    <Admin appLayout={MyLayout} dataProvider={dataProvider} theme={useTheme()}>
+    <Admin layout={Layout} dataProvider={dataProvider} theme={useTheme()}>
         <Resource name="Settings" list={Dashboard} />
         <Resource
             name="queryapis"
             label="Query APIs"
             options={{ label: 'Query APIs' }}
-            list={QueryAPIList}
-            show={QueryAPIShow}
+            list={QueryAPIsList}
```
garethsb-sony: Lack of plural bothered you as well then :laughing:

Development/src/pages/flows/FlowsShow.js
```diff
@@ -0,0 +1,180 @@
...
+import JsonIcon from '../../components/JsonIcon';
+import MapTags from '../../components/TagsField';
+import TAIField from '../../components/TAIField';
+import QueryVersion from '../../components/QueryVersion';
+import ChipConditionalLabel from '../../components/ChipConditionalLabel';
+
+const cookies = new Cookies();
+
+const FlowsTitle = ({ ...props }) => {
+    return (
```
garethsb-sony: Why is this one using `props` not `record` like (all?) the others?

Development/src/pages/devices/DevicesShow.js
```diff
@@ -0,0 +1,133 @@
...
+                )}
+                <ReferenceField
+                    label="Node"
+                    source="node_id"
+                    reference="nodes"
+                    link="show"
+                >
+                    <ChipConditionalLabel source="label" />
+                </ReferenceField>
+                <ReferenceArrayField source="receivers" reference="receivers">
```
garethsb-sony: Are all the attributes that have gone from these fields actually the defaults, then?

Development/src/dataProvider.js
```diff
@@ -212,16 +236,16 @@ const convertDataProviderRequestToHTTP = (type, resource, params) => {
         }
         case GET_MANY_REFERENCE: {
             let total_query;
-            if (params.target !== '' && params[params.source] !== '') {
+            if (params.target !== '' && params.id !== '') {
```
garethsb-sony: Can you talk me through this change, please?

Development/src/App.js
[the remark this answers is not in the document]
rufusu-sony: Although I have found plugins that sorts imports, I don't think it's as simple as you may still want some form of ordering. For example common convention is to have react and core react modules such as react-router at the top, package imports secondly and then local imports. We'd ideally want three distinct sections, alphabetised. Maybe it is worth exploring if such a third party plugin would exist?

Development/src/pages/queryapis/QueryAPIsList.js
rufusu-sony: As this is the outlier however, maybe it would be worth making a separate title function the exception rather than the norm? It's not hurting anyone how it is, just a personal distaste for unnecessary repetition.

Development/src/pages/devices/DevicesList.js
rufusu-sony: Correct, although as you have pointed out this seems have been used interchangeably with the explicit `<Fragment>` syntax. I think it may be worth being more explicit as at first glance `<>` can look like a mistake!

Development/src/pages/logs.js
rufusu-sony: Unfortunately I was unable to find a way to make Clear All Filters correctly function in the time frame I had. This is because the `<FilterField>` component stores it's filter value in it's own state. As child props cannot be modified from a parent component, the only way to refresh the filter components is to reload the whole page. In the future, filter state will have to be managed by the Parent for this to work

garethsb-sony: I don't think it's necessary, honestly.

Development/src/pages/receivers/ReceiversShow.js
rufusu-sony: Refresh is now a react hook provided by react-admin. As MakeConnections is not a react functional component, we cannot use hooks within it.

Development/src/App.js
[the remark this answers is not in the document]
garethsb-sony: Something like https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md maybe?

Development/src/dataProvider.js
rufusu-sony: The major version bump broke `<ReferenceManyField>` components, params.source was no longer passed to the dataProvider and params.id seemed a fitting replacement.

garethsb-sony: You used to be able to override which of the current resource's properties was the source to be looked up: https://marmelab.com/react-admin/Fields.html#referencemanyfield
Is that not the case any longer?

Development/src/pages/devices/DevicesShow.js
garethsb-sony: We realised that it's a bit awkward to get the resource type, so leaving this for a future PR.

Development/src/dataProvider.js
garethsb-sony: Looks like the react-admin framework maps source to id, so we don't need to do it in the dataProvider any longer.

Development/src/pages/nodes/NodesShow.js
garethsb-sony: Done

PR #20, opened 2019-12-04
Move cursor pagination to the UI layer and minor cosmetic changes

rufusu-sony:
- Cursor pagination URLs are now passed to the UI layer.
- This removes the work around requiring random numbers to be generated to cause an update.
- The `<PaginationButton>` component has been refactored
- Fix floating tabs in Chrome
- Add spacing and raise transport parameter cards

### DIFF
--- a/Development/src/pages/devices/DevicesList.js
+++ b/Development/src/pages/devices/DevicesList.js
@@ -8,28 +8,26 @@ import {
     TableHead,
     TableRow,
 } from '@material-ui/core';
-import { Loading, ShowButton, Title } from 'react-admin';
+import { Loading, ShowButton, Title, useRefresh } from 'react-admin';
 import FilterField from '../../components/FilterField';
 import PaginationButton from '../../components/PaginationButton';
 import ListActions from '../../components/ListActions';
 import useGetList from '../../components/useGetList';
 
 const DevicesList = props => {
+    const refresh = useRefresh();
     const [filter, setFilter] = useState({});
-    const [paginationCursor, setPaginationCursor] = useState(null);
-    // As the paginationCursor variable has not changed we need to force an update
-    const [seed, setSeed] = useState(Math.random());
-
-    const { data, loaded, url } = useGetList({
+    const [paginationURL, setPaginationURL] = useState(null);
+    const { data, loaded, pagination, url } = useGetList({
         ...props,
         filter,
-        paginationCursor,
-        seed,
+        paginationURL,
     });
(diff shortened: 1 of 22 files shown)


rufusu-sony:
This is fully functional however there is still duplicated code. Would it be worth refactoring List views to higher level components for this pull request?
Ultimately what I've been thinking of is the same level of abstraction found in the react-admin `<DataGrid>` and `<List>` components where we can build views like so:
```jsx
const SendersList = props => (
    <List {...props}>
        <Datagrid>
            <TestField source="label" />
            <TextField source="transport" />
            <TextField source="subscription.active" />
        </Datagrid>
    </List>
);
```

garethsb-sony:
Looking good, and yes, I think that should be ultimate goal, but should be second step. Let's review this in detail early next week. 👍

Development/src/pages/devices/DevicesList.js
```diff
@@ -98,10 +101,10 @@ const DevicesList = props => {
                         </TableBody>
                     </Table>
                     <br />
-                    <PaginationButton label="FIRST" nextPage={nextPage} />
-                    <PaginationButton label="PREV" nextPage={nextPage} />
-                    <PaginationButton label="NEXT" nextPage={nextPage} />
-                    <PaginationButton label="LAST" nextPage={nextPage} />
+                    <ListPaginationButton label="first" />
```
garethsb-sony: I'd like to see this as `<ListPaginationButton label="First" rel="first"/>` with `label` defaulted from `rel` in the same way as other React-Admin components default their label from `source` or `name` props. Is that doable quickly?

garethsb-sony: And I think you're right that defining a `<Pagination/>` component that created all four (or two, if no first/last cursors provided) buttons would be a good step. Even if we don't fully fit back into the React-Admin model, https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/list/Pagination.js could provide inspiration for the interface?

rufusu-sony:
I also have a couple very minor changes made on my [Development Branch](https://github.com/rufusu-sony/nmos-js/commits/development)
This includes:
- Fixing the QueryAPI connect button as it is non functional
- Improving docker build times with a .dockerignore file

Is it a good idea to bundle these with this pull request or would you like a follow up request after this is merged?

garethsb-sony:
Thanks, Rufus. I incorporated the other changes from dev. :+1:

PR #24, opened 2019-12-13
List view active toggles and SDP patching

rufusu-sony:
- Add toggle switches to the Active column on List views.
- If the sender has an SDP file, patch it along with the transport_params.
- Unneeded fetch requests when fetching endpoints are aborted when the required data is retrieved or an error occurs.
- Replaced function() with ES6 arrow syntax for consistency.
- Sender labels are now chip fields and the device column has been removed


Development/src/pages/receivers/ReceiversShow.js
```diff
@@ -424,14 +446,18 @@ const ConnectionManagementTab = ({
...
-                                    <ShowButton
-                                        style={{
-                                            textTransform: 'none',
-                                        }}
-                                        basePath="/senders"
-                                        record={item}
-                                        label={item.label}
-                                    />
+                                    <Link
+                                        to={`${linkToRecord(
```
garethsb-sony: Would a `ReferenceField` work here?

Development/src/pages/receivers/ReceiversShow.js
```diff
@@ -83,29 +93,28 @@ const ReceiversShow = props => {
...
+                                }
+                            />
+                        ))}
                         <Tab
                             label="Connect"
                             value={`${props.match.url}/connect`}
                             component={Link}
                             to={`${props.basePath}/${props.id}/show/connect`}
-                            disabled={!get(controllerProps.record, '$active')}
+                            disabled={
```
garethsb-sony: Can't see that it matters either way, but I think using `'$staged'` makes more sense logically cos that's what the connect tab uses, isn't it?

Development/src/components/makeConnection.js
```diff
@@ -120,94 +203,15 @@ const makeConnection = (senderID, receiverID, endpoint) => {
...
             })
-            .then(params =>
-                dataProvider('UPDATE', 'receivers', params, {
-                    onSuccess: { refresh: true },
-                })
-            )
+            .then(params => dataProvider('UPDATE', 'receivers', params))
             .then(() => resolve())
-            .catch(error => reject(error.message));
+            .catch(error => reject(error));
```
garethsb-sony: Hmm, where has the use of `message` ended up?

Development/src/components/ActiveField.js
```diff
@@ -0,0 +1,86 @@
...
+        get(record, 'subscription.active')
+    );
+
+    const handleChange = (record, resource) => {
+        toggleMasterEnable(record, resource)
+            .then(() => setChecked(!checked))
+            .catch(error => notify(error.toString(), 'warning'));
+    };
+
+    useEffect(() => {
```
garethsb-sony: Would it be worth a comment here to say why we're doing this rather than refresh?

garethsb-sony:
What happens in the Receivers and Senders list view if you try to use the Active toggle when the Receiver/Sender doesn't support IS-05? Would it be possible to disable the toggle switch in that case, please?

Development/src/pages/receivers/ReceiversShow.js
rufusu-sony: Difference is ReferenceField will make another network request although it would look more simple.

garethsb-sony: Ah, yes, that makes sense, and is a good reason. Please add that as a comment here.

Development/src/components/makeConnection.js
rufusu-sony: As the error may be network related and would therefore have an error.debug field, I wanted to wait until we handle the error before using the string form. The error class additionally has a toString() method which we should be using.

rufusu-sony:
What is the best way to tell if IS-05 is enabled from the IS-04 record? Will I have to get the device of each receiver and look at the controls?

Development/src/components/makeConnection.js
garethsb-sony: Sure, but in other places I can only see handling of the NMOS form with `debug`?

Development/src/components/ActiveField.js
garethsb-sony: That comment doesn't make sense to me, but that may be unfamiliarity with React terminology... what do 'refresh' and 'sent' mean in this instance?

Development/src/components/ActiveField.js
```diff
@@ -60,10 +60,11 @@ const ActiveField = ({ className, source, record = {}, resource, ...rest }) => {
 
     const handleChange = (record, resource) => {
         toggleMasterEnable(record, resource)
-            .then(() => setChecked(!checked))
+            .then(({ data }) => setChecked(get(data, 'master_enable')))
```
garethsb-sony: So `toggleMasterEnable` returns the new IS-05 /active endpoint data does it?

rufusu-sony: Yes, thought it best to base the status of the switch off the response from the IS-05 endpoint rather than simply toggling it

Development/src/components/ActiveField.js
rufusu-sony: Let me reword this, then I'll force push over this commit with the other changes.
Does this make more sense?
When the page refresh button is pressed, the ActiveField will receive a new record prop. When this happens we should update the state of the switch to reflect the newest IS-04 data

Development/src/components/makeConnection.js
rufusu-sony: Apologies but I don't quite follow, where specifically are you referencing?
An oddity with the format of errors is they may not always have a debug field. Unfortunately using the error.toString() method does not include the debug message in the returned string. This is why I have chosen to handle errors with a debug property separately in some locations, as the debug property includes useful information regarding schema validation etc..

Development/src/components/ActiveField.js
garethsb-sony: Oh, hmm, so this is about page refresh? I kinda assumed that the individual components got torn down at that point, and the table page would be rebuilt from a new response, so I didn't see why there would be special handling for the `ActiveField`. I guess I was wrong, and because we've allowed the component's state to be tweaked, we need to 'clear' that?

Development/src/components/makeConnection.js
garethsb-sony: All the error handling I could see, assumes the NMOS JSON code/message/debug props. So when an error doesn't have those things, what happens? Maybe I am looking at two different levels of error handling.

garethsb-sony:
> What is the best way to tell if IS-05 is enabled from the IS-04 record? Will I have to get the device of each receiver and look at the controls?

If by 'enabled' you mean does it have IS-05, yes, I suppose you'd have to get the Device.controls. I guess I was thinking it was like the decision about whether to show the Staged/Active/Connect tabs, but it seems wrong here when showing the list view to do that for every row, doesn't it?

Perhaps for now, we take the same optimistic approach we did for the Activate/Stage buttons in the Connect tab, and show the toggle enabled, and report error if it doesn't work for whatever reason?

Development/src/components/makeConnection.js
rufusu-sony: Currently if the error does not have the body property containing the debug information, I am just using error.toString()
```javascript
.catch(error => {
    if (error && error.hasOwnProperty('body'))
        notify(
            get(error.body, 'error') +
                ' - ' +
                get(error.body, 'code') +
                ' - ' +
                get(error.body, 'debug'),
            'warning'
        );
    notify(error.toString(), 'warning');
});
```
I believe the ideal solution to this problem would be to rewrite the data provider layer to use our own fetch instead of the react-admin fetchJson. This way we could create our HttpError class with a toString() method that returns the string we would like. Is this work you think would be good to undertake today?

garethsb-sony: No :-)
That makes sense. Thanks!

Development/src/components/ActiveField.js
rufusu-sony: To prevent a jerky page movement I'm updating all components in place yes. If I wanted to fully reset the components I could tear them down as you were saying. This is functionally equivalent as we reset the component state to it's initial value.
The specific case I was targeting was if an external factor has altered the IS-04 data such as another client.

garethsb-sony: And by 'initial value' you mean "new value, but without being masked by the 'local' change made by toggling the switch"?

rufusu-sony:
For now I will add a simple error message if the connection API could not be reached

Development/src/components/ActiveField.js
rufusu-sony: Yes, that's a better way of putting it

Development/src/components/ActiveField.js
```diff
@@ -33,21 +33,24 @@ const toggleMasterEnable = (record, resource) => {
             id: record.id,
         })
             .then(({ data }) => {
+                if (!data.hasOwnProperty('$staged')) {
+                    throw new Error('Could not reach connection API');
```
garethsb-sony: connection -> Connection

And there are two possibilities, right? Either the IS-04 Device simply doesn't list any Connection API endpoints in its `controls`, or none of them responded? Does 'reach' cover both of those? 🤔

Development/src/components/ActiveField.js
garethsb-sony: Perfect

Development/src/components/ActiveField.js
rufusu-sony: Would 'No connection API found' be a suitable message?

garethsb-sony: Yes - with the correct capitalization. 😉

PR #25, opened 2019-12-17
Fix console spam and reformat QueryAPI pages

rufusu-sony:
- Don't attempt to translate objects
- Sanitize props passed to react DOM elements
- Remove redundant refresh prop on receivers list view
- Separate `<Table>` from `<SimpleShowLayout>` components with a <Fragment>
  to prevent props being passed
- Remove redundant `<div>` from card views
- A view raw button has been added to the QueryAPI list and show views
- The connect button has been refactored and a selection menu added
- The QueryAPI show button has been combined with the label
- Padding has been added to the list view


garethsb-sony:
This is really nice, thank you, Rufus.

PR #27, opened 2019-12-20
New filtering UI

rufusu-sony:
- List views have been updated to use a dynamic system where filters can be added and removed.
- Allows filtering fields not displayed


garethsb-sony:
Removing a filter by clicking the close box doesn't seem to remove it from the underlying filter?

garethsb-sony:
In Chrome, the filter icon and 'Add Filter' don't quite fit in the button, and when many filters are being displayed, the last one also overlaps the left part of the filter icon.

garethsb-sony:
When you add a filter, would it be nice if the field got the focus, so you could start typing straight away? (Don't know if keyboard shortcuts apply to the toggle switch for booleans?)

garethsb-sony:
Something's not quite right with the 'Level' filter on the Logs list. When I add the filter, all rows disappear... probably it should have no effect if it's empty? And when I then type '-40', I briefly see rows appear but then they all disappear again and don't come back.

PR #29, opened 2020-01-07
Use new concatUrl function "to avoid cases of double slashes"

### LINKED ISSUE #28: Additional / (slash) character in HTTP requests
I've noticed whilst attempting to connect a sender to a receiver (Immediate Activation) that there are additional "/" characters in the HTTP Requests. To be clear, I'm not referring to trailing slashes (as previously raised in other tickets) but duplicate "/" in the middle of a URL.

When pressing "Activate", a GET request is sent to the sender and a little later the receiver - however a pcap reveals that there is an additional "/" after the API version number - resulting in HTTP 404 error message. For example:
GET /x-nmos/connection/v1.0//single/receivers/70b3d504-2d09-1000-8000-000000000099

image

Manually performing this request (without the extra slash) is successful:
GET /x-nmos/connection/v1.0/single/receivers/70b3d504-2d09-1000-8000-000000000099

Apologies if I'm missing something obvious here, but so far as I can tell this is a bug?

garethsb-sony:
… as per guidance in IS-04 and IS-05 (see [IS-04 Client Behaviour](https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.3.x/docs/2.0.%20APIs.md#client-behaviour) and same info in [IS-05 Discovery](https://github.com/AMWA-TV/nmos-device-connection-management/blob/v1.1.x/docs/3.1.%20Interoperability%20-%20NMOS%20IS-04.md#discovery)).

Will resolve #28. Thanks, @andrewjohnwatts, for the original report.

### DIFF
--- a/Development/src/components/ConnectionShowActions.js
+++ b/Development/src/components/ConnectionShowActions.js
@@ -2,12 +2,10 @@ import React from 'react';
 import NavLink from 'react-router-dom/NavLink';
 import { Button, ListButton, TopToolbar, useGetOne } from 'react-admin';
 import get from 'lodash/get';
-import Cookies from 'universal-cookie';
 import EditIcon from '@material-ui/icons/Edit';
 import JsonIcon from '../icons/JsonIcon';
 import { useTheme } from '@material-ui/styles';
-
-const cookies = new Cookies();
+import { concatUrl, resourceUrl } from '../dataProvider';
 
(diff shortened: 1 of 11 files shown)


garethsb-sony:
@rufusu-sony, review please... I reckon there's a bit more we can do here, to tidy up new RawButton vs. existing ListActions and ConnectionShowActions?

PR #32, opened 2020-01-24
Add creating and deleting subscriptions

rufusu-sony:
- A new screen for creating a subscriptions as
  well as delete buttons on list and show views.

### DIFF
--- a/Development/src/dataProvider.js
+++ b/Development/src/dataProvider.js
@@ -147,7 +147,10 @@ const convertDataProviderRequestToHTTP = (type, resource, params) => {
                     let parsedValue = encodeURIComponent(value);
                     parsedValue = parsedValue.split('%2C'); //splits comma separated values
                     for (let i = 0; i < parsedValue.length; i++) {
-                        if (key === 'level' || typeof value === 'boolean') {
+                        if (
+                            ['max_update_rate_ms', 'level'].includes(key) ||
+                            typeof value === 'boolean'
+                        ) {
                             if (value !== '')
                                 matchParams.push(
                                     'eq(' + key + ',' + parsedValue[i] + ')'
--- a/Development/src/dataProvider.js
+++ b/Development/src/dataProvider.js
@@ -279,10 +279,22 @@ const convertDataProviderRequestToHTTP = (type, resource, params) => {
                 url: concatUrl(params.data.$connectionAPI, '/staged'),
                 options: options,
             };
-        case CREATE:
-            return '';
+        case CREATE: {
+            const url = resourceUrl(resource, `?${resource}`);
+            const options = {
+                method: 'POST',
+                body: JSON.stringify(params.data),
+            };
+            return {
+                url,
+                options: options,
+            };
+        }
         case DELETE:
-            return '';
+            return {
+                url: resourceUrl(resource, `/${params.id}`),
+                options: { method: 'DELETE' },
+            };
         default:
             //not expected to be used
             return '';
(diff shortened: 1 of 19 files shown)


Development/src/dataProvider.js
```diff
@@ -279,10 +279,22 @@ const convertDataProviderRequestToHTTP = (type, resource, params) => {
                 url: concatUrl(params.data.$connectionAPI, '/staged'),
                 options: options,
             };
-        case CREATE:
-            return '';
+        case CREATE: {
+            const url = resourceUrl(resource, `?${resource}`);
```
garethsb-sony: What's the `?${resource}` for?

Development/src/pages/subscriptions/SubscriptionsCreate.js
```diff
@@ -0,0 +1,39 @@
...
+const SubscriptionsCreate = props => (
+    <Create actions={<SubscriptionsCreateActions />} {...props}>
+        <SimpleForm redirect="show">
+            <TextInput source="resource_path" label="Resource Path" />
+            <NumberInput
+                source="max_update_rate_ms"
+                label="Max Update Rate (ms)"
+            />
+            <ObjectInput source="params" />
+            <BooleanInput source="persist" initialValue={false} />
```
garethsb-sony: Probably ought to default it to `true` actually.

Development/src/pages/subscriptions/SubscriptionsCreate.js
```diff
@@ -0,0 +1,39 @@
...
+    <TopToolbar>
+        {data ? <RawButton record={data} resource={resource} /> : null}
+        <ListButton basePath={basePath} />
+    </TopToolbar>
+);
+
+const SubscriptionsCreate = props => (
+    <Create actions={<SubscriptionsCreateActions />} {...props}>
+        <SimpleForm redirect="show">
+            <TextInput source="resource_path" label="Resource Path" />
```
garethsb-sony: Can we add a set of allowed values for `resource_path`? Maybe using [SelectInput](https://marmelab.com/react-admin/Inputs.html#selectinput)

Development/src/pages/subscriptions/SubscriptionsCreate.js
```diff
@@ -0,0 +1,39 @@
...
+            <NumberInput
+                source="max_update_rate_ms"
+                label="Max Update Rate (ms)"
+            />
+            <ObjectInput source="params" />
+            <BooleanInput source="persist" initialValue={false} />
+            {QueryVersion() >= 'v1.3' && (
+                <BooleanInput source="authorization" />
```
garethsb-sony: Since we're not letting the user choose a value for `secure`, we should probably remove `authorization` from the form too.

garethsb-sony:
Changes look good, only a couple of trivial questions. Will test!

garethsb-sony:
Last problem I have is that the refresh/redirect back to the list page isn't removing the deleted Subscription from the list. Related to this [old issue](https://github.com/marmelab/react-admin/issues/2425) maybe?

garethsb-sony:
OK, commit fa9780c resolves that issue for me, and everything is now working nicely, but there's probably a more elegant solution. @rufusu-sony, please would you review?

garethsb-sony:
Added a couple more commits, but wondering whether a better solution for the Logs `level` filter and Subscriptions `max_update_rate_ms` filter would be to add a `NumberFilter` using a `NumberField` but otherwise just like the `StringFilter`?

PR #34, opened 2020-02-05
Change AutoRefresh behaviour

rufusu-sony:
- Only refresh on list and show views
- Reset counter on page change


rufusu-sony:
The best way I have found to discover if an edit view is being used is to look at the number of URL segments. This may not be optimal but is necessary as React-Admin doesn't include an 'edit' keyword in the URL for edit pages.
If a better way is found to detect edit pages, this can be easily incorporated.

garethsb-sony:
Thank you, Rufus!

PR #37, opened 2020-03-24
When making a connection, force the dataProvider to include rtp_enabled

garethsb-sony:
The spec doesn't define any means to indicate the active status of the Sender's legs in the SDP file, and shows `rtp_enabled` in the Receiver's legs is just set to true or false based on the number of legs in the SDP file, so making this change results in least surprise for users.


rufusu-sony:
The change makes sense to me. Regarding programming convention, does it make sense to modify and return a variable called previousData as it may suggest nothing has been modified?
Otherwise looks good to me.

PR #40, opened 2020-04-14
Fix connect tab removing filter on refresh

rufusu-sony:
- Connect tab has been refactored into a separate file as the logic had nothing in common with the receiver show view.
- To prevent unnecessary recreations of the connect tab, the component is stored as state and only recreated when the receiver data is changed.
- The render prop has been removed from the tab Route components.
- The 'Pagination' cookie was used to communicate the paging capabilities of an API to the list view. Storing this as a cookie caused bugs when multiple requests for different resources are made, ie the connections page.
- The enabled pagination features are now discerned from keys of the pagination prop which was already being passed to the list view.

### DIFF
--- a/Development/src/components/PaginationButtons.js
+++ b/Development/src/components/PaginationButtons.js
@@ -6,9 +6,8 @@ import {
     LastPage,
 } from '@material-ui/icons';
 import { Button } from '@material-ui/core';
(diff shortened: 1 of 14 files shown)


garethsb-sony:
Much nicer (as well as fixing the bug!). Thanks, @rufusu-sony !

PR #41, opened 2020-05-01
Autofill filters

rufusu-sony:
- Add a defaultFilter and defaultValue props to FilterPanel.js to allow filters to be automatically populated

### DIFF
--- a/Development/src/components/FilterPanel.js
+++ b/Development/src/components/FilterPanel.js
@@ -40,10 +40,12 @@ export const BooleanFilter = ({
     autoFocus,
 }) => {
     const [checked, setChecked] = useState(() => {
-        if (defaultValue != null) {
+        if (filter[source] != null) {
+            return !!filter[source];
+        } else if (defaultValue != null) {
             return defaultValue;
         } else {
-            return !!filter[source];
+            return false;
         }
     });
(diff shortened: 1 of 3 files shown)


Development/src/components/FilterPanel.js
```diff
@@ -74,14 +81,23 @@ export const BooleanFilter = ({
...
     source,
     label,
     filter,
     setFilter,
     autoFocus,
     ...props
 }) => {
-    const [value, setValue] = useState(filter[source] ? filter[source] : '');
+    const [value, setValue] = useState(() => {
+        if (defaultValue) {
```
garethsb-sony, line 93: @RufusUtt, it seems backwards to me to prefer the `defaultState`? I thought this should be:
```javascript
if (filter[source]) {
    return filter[source];
} else if (defaultValue) {
    return defaultValue;
} else {
    return '';
}
```
which I think can be written more succinctly as:
```javascript
return filter[source] || defaultValue || '';
```

But I've probably missed something!

garethsb-sony: I've made this tweak (apart from using special treatment for `NaN`).
I also fixed `RateFilter` which I'd broken while refactoring the original PR 🙈

PR #49, opened 2021-01-05
IS-08 channel mapping

garethsb-sony:


Development/src/components/CollapseButton.js
```diff
@@ -0,0 +1,15 @@
+import { IconButton } from '@material-ui/core';
+import KeyboardArrowDownIcon from '@material-ui/icons/KeyboardArrowDown';
+import KeyboardArrowUpIcon from '@material-ui/icons/KeyboardArrowUp';
+
+const CollapseButton = ({ onClick, isExpand }) => (
+    <IconButton
+        size="small"
+        title={isExpand ? 'Hide channels' : 'View channels'}
```
garethsb-sony, line 8: Collapse Button da a generic name, but this is specific to channels. Maybe pull the title word out?

Development/src/components/MappingButton.js
```diff
@@ -0,0 +1,11 @@
+import { IconButton } from '@material-ui/core';
+import CheckCircleOutlineIcon from '@material-ui/icons/CheckCircleOutline';
+import RadioButtonUncheckedIcon from '@material-ui/icons/RadioButtonUnchecked';
+
+const MappingButton = ({ disabeld, onClick, isMapped }) => (
```
garethsb-sony, line 5: Typo

Development/src/components/MappingShowActions.js
```diff
@@ -0,0 +1,61 @@
...
+                </Button>
+            ) : null}
+            <ListButton
+                label={'List'}
+                title={'Return to ' + basePath}
+                basePath={basePath}
+            />
+            {get(data, '$channelMappingAPI') != null ? (
+                <Button
+                    label={'Edit'}
```
garethsb-sony, line 52: Seeing this makes me realise that 'Edit' in the top toolbar for Device, and Receiver and Sender, is fine right now as there's only one editable tab/API in each case (IS-08 and IS-05 Staged), but if we end up with other `"services" ` to support we'll need a rethink.

Development/src/dataProvider.js
```diff
@@ -674,6 +674,79 @@ const getEndpoints = (addresses, resource, id) => {
     });
 };
 
+const getChannelMappingEndPoints = (addresses, endpoints) => {
+    // Looking for the first promise to succeed or all to fail
```
garethsb-sony, line 678: Let's pull these out ultimately

Development/src/dataProvider.js
```diff
@@ -674,6 +674,79 @@ const getEndpoints = (addresses, resource, id) => {
     });
 };
 
+const getChannelMappingEndPoints = (addresses, endpoints) => {
```
garethsb-sony, line 677: Let's rename `getEndpoints` to `getConnectionEndPoints` or something!

Development/src/dataProvider.js
```diff
@@ -786,6 +859,60 @@ const convertHTTPResponseToDataProvider = async (
                     });
                 }
             }
+            if (resource === 'devices') {
+                let deviceJSONData = await fetch(
```
garethsb-sony, line 863: Isn't `json` the IS-04 Device data, so maybe no need to refetch?

Development/src/pages/devices/DevicesShow.js
```diff
@@ -32,12 +40,72 @@ export const DevicesShow = props => {
 };
 
 const DevicesShowView = props => {
+    const { record } = useRecordContext();
+    const [useIS08API, setUseIS08API] = useState(false);
```
garethsb-sony, line 44: Although it would be more verbose, I'd prefer `useChannelMappingAPI`, since AMWA IS-xx specs sometimes define multiple APIs.

Development/src/pages/devices/DevicesShow.js
```diff
@@ -32,12 +40,72 @@ export const DevicesShow = props => {
...
+                        indicatorColor="primary"
+                        textColor="primary"
+                    >
+                        <Tab
+                            label="Summary"
+                            value={`${props.match.url}`}
+                            component={Link}
+                            to={`${props.basePath}/${props.id}/show/`}
+                        />
+                        {['Active_Matrix', 'Staged_Matrix'].map(label => (
```
garethsb-sony, line 78: Although we generally try to stick to the NMOS terminology rather than user friendly terms, in this case I agree it seems completely correct to use *Matrix* rather than *Map*, especially as you are rendering the map as a matrix! 🎉😁

Development/src/pages/devices/DevicesShow.js
```diff
@@ -107,4 +175,22 @@ const DevicesShowView = props => {
     );
 };
 
+const ShowActiveMatrixTab = ({ deviceData, ...props }) => {
+    if (!get(deviceData, `$active.map`)) return <Loading />;
+    return (
+        <ShowView {...props} title={<ResourceTitle />} actions={<Fragment />}>
+            <Card>
+                <Title title={' Active Matrix'} />
```
garethsb-sony, line 183: Is the space at the beginning needed?

Development/src/pages/devices/ChannelMappingMatrix.js
```diff
@@ -0,0 +1,514 @@
...
+
+const get_output_caps = (output_item, io) => {
+    return (
+        <>
+            <Typography variant="body1">{'Description:'}</Typography>
+            {output_item.properties.description}
+            <Typography variant="body1">{'Routable inputs:'}</Typography>
+            {output_item.caps.routable_inputs !== null
+                ? output_item.caps.routable_inputs
+                      .map(input_name => {
```
garethsb-sony, line 29: How about `input_id` for this to distinguish the ID from the human - readable `properties.name`?

Development/src/pages/devices/ChannelMappingMatrix.js
```diff
@@ -0,0 +1,514 @@
...
+                          return input_name === null
+                              ? 'MUTE'
```
garethsb-sony, line 31: A place to think about NMOS terminology vs user friendly terms 🤔 MUTE vs unrouted vs... not sure!

garethsb-sony: (We'd often use the ✖️ icon for this kind of `null` but that doesn't play nicely as part of the list of inputs I expect)

Development/src/pages/devices/ChannelMappingMatrix.js
```diff
@@ -0,0 +1,514 @@
...
+                              : get(io, `inputs.${input_name}.properties.name`);
+                      })
+                      .join(', ')
+                : 'null'}
```
garethsb-sony, line 35: This `null` is "allows any input to be routed", need to think best way to indicate that.

Development/src/components/CollapseButton.js
```diff
@@ -0,0 +1,15 @@
+import { IconButton } from '@material-ui/core';
+import KeyboardArrowDownIcon from '@material-ui/icons/KeyboardArrowDown';
+import KeyboardArrowUpIcon from '@material-ui/icons/KeyboardArrowUp';
+
+const CollapseButton = ({ onClick, isExpand }) => (
```
garethsb-sony, line 5: Maybe name the param `isExpanded` unless there's precedent?

Development/src/pages/devices/ChannelMappingMatrix.js
```diff
@@ -0,0 +1,514 @@
...
+                            </div>
+                        </ReferenceManyField>
+                    </SingleFieldList>
+                </ReferenceManyField>
+            </div>
+        </>
+    );
+};
+
+const OutputSourceAssociation = ({ outputs, isExpend }) =>
```
garethsb-sony, line 121: Maybe `isExpanded`?

Development/src/pages/devices/ChannelMappingMatrix.js
garethsb-sony: I think there are a few other places where `name` (and `nmos_name`) is used to mean the input or output ID in the *io* object. Could we make all of these use `id` consistently?

Development/src/pages/devices/ChannelMappingMatrix.js
```diff
@@ -0,0 +1,514 @@
...
+                </ReferenceManyField>
+            </div>
+            <Typography variant="body1">{'Senders:'} </Typography>
+            <div
+                style={{
+                    margin: 2,
+                    padding: 1,
+                }}
+            >
+                <ReferenceManyField
```
garethsb-sony, line 86: I wonder how many API calls the nested `ReferenceManyField`s result in, and whether it's worth trying to be more efficient.

garethsb-sony:
This is looking very good, @orfar1994! All comments so far are very minor.

Development/src/pages/devices/DevicesShow.js
garethsb-sony: Agreed in call today to leave page title unchanged on the tabs

Development/src/pages/devices/ChannelMappingMatrix.js
garethsb-sony: We discussed in call today, maybe leaving as MUTE (or MUTED better?), though the IS-08 spec doesn't use that word, instead "unrouted" or "parked", "silence", or just `null`.

> An Output channel that does not have an Input channel routed to it is said to be "unrouted" (sometimes known as "parked")
([link](https://amwa-tv.github.io/nmos-audio-channel-mapping/tags/v1.0.1/docs/1.0._Overview.html#terminology))

> If no Input channel is being routed to an Output channel, both `input` and `channel_index` MUST be set to `null`. Where an Output does not have a channel routed to it the Device MUST interpret this as silence, and produce a silent signal of appropriate bit depth and bit rate from that output.
([link](https://amwa-tv.github.io/nmos-audio-channel-mapping/tags/v1.0.1/docs/4.0._Behaviour.html#the-active-map))

Development/src/pages/devices/ChannelMappingMatrix.js
garethsb-sony: In the call we discussed writing this as ANY. I found the spec uses the phrase "no restrictions", so maybe that's also a possibility:

>If no such restrictions exist, the `routable_inputs` field MUST be set to `null`.
([link](https://amwa-tv.github.io/nmos-audio-channel-mapping/tags/v1.0.1/docs/4.0._Behaviour.html#output-routing-constraints))

Development/src/pages/devices/ChannelMappingMatrix.js
garethsb-sony: Looked good on the call today! Worth keeping an eye on the browser developer tools console and network tabs though, to keep the number of API calls and console logging minimized.

I will investigate whether the `paging.limit=1000` that @rhastie spotted will cause any problems.

Development/src/pages/devices/DevicesShow.js
garethsb-sony: In the call we decided to make the URL keys be `active_map` and `staged_map` and the UI show 'Active Map' and 'Staged Map'.

Development/src/pages/devices/ChannelMappingMatrix.js
```diff
@@ -0,0 +1,514 @@
...
+            ) : null}
+        </TableCell>
+    ));
+
+const InputSourceAssociation = ({ isRowExpanded, input_item }) => (
+    <TableCell
+        align="center"
+        rowSpan={isRowExpanded ? input_item.channels.length + 1 : 1}
+    >
+        {input_item.parent.type === 'source' ? (
```
garethsb-sony, line 156: In the call today, we discussed adding a tooltip here with "Type: Source" or "Type: Receiver".

Development/src/pages/devices/ChannelMappingMatrix.js
```diff
@@ -0,0 +1,514 @@
...
+                record={input_item}
+                basePath="/receivers"
+                label="Receivers"
+                source="parent.id"
+                reference="receivers"
+                link="show"
+            >
+                <ChipConditionalLabel source="label" />
+            </ReferenceField>
+        ) : null}
```
garethsb-sony, line 178: We discussed also whether blank cell was the right thing here. I wonder whether a simple text placeholder (instead of the chip) saying "No parent" might be nice?

Development/src/pages/devices/ChannelMappingMatrix.js
```diff
@@ -0,0 +1,514 @@
...
+    </TableCell>
+);
+
+const FillEmptyCellsForCollapseRow = ({ outputs, isColExpanded }) =>
```
garethsb-sony, line 182: `EmptyCellsForCollapsedRow` maybe

Development/src/pages/devices/ChannelMappingMatrix.js
```diff
@@ -0,0 +1,514 @@
...
+                {outputs.map(([output_nmos_name, output_item]) => {
+                    return isColExpanded(output_nmos_name) ? (
+                        output_item.channels.map((output_channel, index) => {
+                            return (
+                                <TableCell align="center">
+                                    <Tooltip
+                                        title={
+                                            <>
+                                                <Typography variant="body1">
+                                                    {'Input Channel: '}
```
garethsb-sony, line 213: In the call today, we discussed that this tooltip should show the Output name as well as the channel, because the column heading might have been scrolled out of view.

Afterwards I realised this also applies to the row heading for the Input, especially for inputs with large numbers of channels. Therefore I think we need Input name as well.

Maybe something simple like 'Input: {input-name} {input-channel-label}' on one line and same for Output on the next line?

Development/src/pages/devices/ChannelMappingMatrix.js
```diff
@@ -0,0 +1,514 @@
...
+                    mapping_disabeld={is_show}
+                    handleMap={handleMap}
+                    isMapped={isMapped}
+                />
+            )}
+        </>
+    ));
+};
+
+const SortByIOName = io_object => {
```
garethsb-sony, line 415: I'm not expert on React coding style guidelines, but this looks like a function not React element, so I'd have expected lowercase first letter, e.g. `sortByIOName` in this case?

Development/src/pages/devices/ChannelMappingMatrix.js
```diff
@@ -0,0 +1,514 @@
...
+                    isMapped={isMapped}
+                    isColExpanded={output => isColExpanded(output)}
+                />
+                <InputsRows
+                    inputs={sorted_inputs}
+                    outputs={sorted_outputs}
+                    isColExpanded={output => isColExpanded(output)}
+                    isRowExpanded={input => isRowExpanded(input)}
+                    handleEpandRow={handleEpandRow}
+                    is_show={is_show}
```
garethsb-sony, line 505: Why the mix between `camelCase` and `snake_case` names here?

Development/src/pages/devices/ChannelMappingMatrix.js
```diff
@@ -0,0 +1,514 @@
...
+                        `${output_name}.${output_channel}.channel_index`
+                    )
+                )
+        );
+    };
+    const io = get(record, '$io');
+    const sorted_outputs = SortByIOName(get(io, `outputs`));
+    const sorted_inputs = SortByIOName(get(io, `inputs`));
+    return (
+        <Table border={1}>
```
garethsb-sony, line 468: What I'm seeing in my browser is a one pixel black border round the whole table, and one pixel black borders vertically between columns, but a one pixel pale grey border horizontally between rows. If we could all of those in the pale grey, I think that would match the look'n'feel of the list views, etc. better?

orfar1994:
This PR include some functionality in order to support IS-08 channel mapping in the controller-
1. "active map" tab with all the fixes from the CR.
2. fetch channel mapping end point in series.

Development/src/pages/devices/ChannelMappingMatrix.js
```diff
@@ -0,0 +1,574 @@
...
+    const [expandedRows, setExpandedRows] = useState([]);
+    const handleEpandRow = inputId => {
+        const currentExpandedRows = expandedRows;
+        const isRowExpanded = currentExpandedRows.includes(inputId);
+        const newExpandedRows = isRowExpanded
+            ? currentExpandedRows.filter(id => id !== inputId)
+            : currentExpandedRows.concat(inputId);
+        setExpandedRows(newExpandedRows);
+    };
+    const handleEpandCol = outputId => {
```
garethsb-sony, line 506: Little typo, `handleExpandCol`

Development/src/pages/devices/ChannelMappingMatrix.js
```diff
@@ -0,0 +1,514 @@
...
+                <ReferenceManyField
+                    record={output_item}
+                    basePath="/flows"
+                    label="Flows"
+                    source="source_id"
+                    reference="flows"
+                    target="source_id"
+                    link="show"
+                >
+                    <div
```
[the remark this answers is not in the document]
garethsb-sony: Nice work on resolving the browser console noise, @orfar1994. I have one concern, about the use of channel labels as unique keys, since I don't believe the IS-08 schema requires that the labels are unique within the input or output. Maybe better to use channel index instead? (I think it's only the most local scope in which React requires the keys to be unique.)

Development/src/pages/devices/ChannelMappingMatrix.js
garethsb-sony: I will commit a change to use 'Unrouted' rather than 'MUTE' for this first PR, and agreed with @rhastie that we can change depending on initial demo feedback.

Development/src/pages/devices/ChannelMappingMatrix.js
garethsb-sony: Similarly, will make this 'No Constraints' for initial merge.

Development/src/pages/devices/ChannelMappingMatrix.js
garethsb-sony: Will commit fix.

Development/src/pages/devices/ChannelMappingMatrix.js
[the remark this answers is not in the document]
orfar1994: React requires the keys to be unique on each map function. what do you think about input ID/output ID/channel index as a unique key?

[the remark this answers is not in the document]
garethsb-sony: I think therefore only necessary to use channel index since each call to `map` only goes over a single input or output's channels. I've pushed this change, see what you think :-)

garethsb-sony:
@orfar1994 - I can merge this today, I'm happy we've resolved all the comments and Rich and I have sorted the admin. 👍 

One thing that I thought would be nice, though I'm not sure how to do, would be if the `CollapseButton` were 'sticky' to the _left_ of the **Output** cell, and to the _top_ of the **Input** cell, so that when they are toggled they don't move very much. Not sure how to do that though!

(Also wondering if CollapseButton ought to have been called `ExpandButton`...)

orfar1994:
I have added a small fix to the row span if we have only 1 channel.
I will try to find a solution for the sticky `CollapseButton`.

garethsb-sony:
> I have added a small fix to the row span if we have only 1 channel.

Does the commit I've just pushed seem simpler by removing the special case for 1 channel?

orfar1994:
the special case for 1 channel is mandatory. with your fix if we have 1 channel and expand the row , we will have external empty row.
image

garethsb-sony:
OK, thanks sorry! I just discovered that behaviour when I tried to apply the same logical structure to the UnroutedRow. I don't understand why it doesn't work, but I've force-pushed back to your commit!

orfar1994:
OK :) regarding the sticky `CollapseButton`, I succeeded align it to top but also the text :(
I couldn't put the text in the center... what do you think? what is better? 
image

garethsb-sony:
That's not ideal is it? Hmm.

I wonder if there's some combination of `position` attributes that would help.

I tried `position: relative` on the table cells and `position: absolute` on the buttons, and got to this:

image

As you can see, not good enough though, because `position: absolute` means the width and height of the button is not taken into account when deciding the size of the table cell or the position of the text, so it can overlap.

orfar1994:
I see... can we merge it and we will dill with that later? I think it's a small issue.

garethsb-sony:
> the special case for 1 channel is mandatory. with your fix if we have 1 channel and expand the row , we will have external empty row.
> image

I found this actually also applies generally, at least to 2 channels, if the name of the Output is longer and the labels of the channels are short:

rowspan

To solve this, I've applied the same fix generally on master.

PR #53, opened 2021-01-13
Is08 channel mapping cr - filter matrix + rename label feature 

orfar1994:
### This PR include 2 features: 
1. **Rename feature -** the user can give new name to the output/input/channel. 
image

it will be saved locally. the user can do it by clicking the output/input/channel name (the name text became button when hover it).
image

when the user clicking the name button.  the text become editable and the user can edit it. 
The user can save it locally be pressing the `DoneIcon`, cancel editing by pressing `ClearIcon` (the name will be the last name that saved) and return to the original name by pressing `DeleteIcon`.

2. **Filter feature -**  the user can filter the matrix by the following 
- Output label (the real name of the output).
- Personal Output Label (the new name that the user give to the output).
- Output channel label (the real name of the output's channel).
- Personal Output channel Label (the new name that the user give to the output's channel).
- Input label (the real name of the input).
- Personal Input Label (the new name that the user give to the Input).
- Input channel label (the real name of the Input's channel).
- Personal Input channel Label (the new name that the user give to the Input's channel).
- Block size (the capability of the input).
- Reordering (the capability of the input).
- Routable inputs (the capability of the output).
- Limit Label Length- the user can cut all the labels in the matrix by this length. 
- Filter Group - the user can add filter group condition (AND or Or), by default it's AND.
image


Development/src/pages/devices/EditMatrixNamesField.js
```diff
@@ -0,0 +1,315 @@
+import React, { Fragment, useEffect, useRef, useState } from 'react';
+import { Button, IconButton, TextField, Tooltip } from '@material-ui/core';
+import get from 'lodash/get';
+import has from 'lodash/has';
+import CreateIcon from '@material-ui/icons/Create';
+import DeleteIcon from '@material-ui/icons/Delete';
+import ClearIcon from '@material-ui/icons/Clear';
+import DoneIcon from '@material-ui/icons/Done';
+import { isEmpty, set } from 'lodash';
```
orfar1994, line 9: need to merge `get` , `has`, `isEmpty` and `set ` to 1 import.

Development/src/pages/devices/EditMatrixNamesField.js
```diff
@@ -0,0 +1,315 @@
...
+
+export const EditableIONameField = ({
```
orfar1994, line 11: I'm not sure about the name...

Development/src/pages/devices/EditMatrixNamesField.js
```diff
@@ -0,0 +1,315 @@
...
+        };
+    }, [autoFocus]);
+
+    const getName = () => {
+        return get(personalNames, `${ioKey}.${source}.name`)
+            ? getLabelByLenght(get(personalNames, `${ioKey}.${source}.name`))
+            : getLabelByLenght(defaultValue);
+    };
+
+    const removeOverrideName = () => {
```
orfar1994, line 51: will change `removeOverrideName ` to `removePersonalName `

Development/src/pages/devices/EditMatrixNamesField.js
```diff
@@ -0,0 +1,315 @@
...
+    label,
+    personalNames,
+    setPersonalNames,
+    autoFocus,
+    getLabelByLenght,
+    title,
+    ioKey,
+    ...props
+}) => {
+    const [displayRenameField, setDisplayRenameField] = useState(false);
```
orfar1994, line 23: maybe change `displayRenameField, setDisplayRenameField` to `displayPersonalField, setPersonalField`

Development/src/pages/devices/EditMatrixNamesField.js
```diff
@@ -0,0 +1,315 @@
...
+            ? getLabelByLenght(get(personalNames, `${ioKey}.${source}.name`))
+            : getLabelByLenght(defaultValue);
+    };
+
+    const removeOverrideName = () => {
+        setDisplayRenameField(false);
+        setDisplayEditIcon(false);
+        setValue(defaultValue);
+        setPersonalNames(f => {
+            let newpersonalNames = { ...f };
```
orfar1994, line 56: `newpersonalNames ` to `newPersonalNames `

Development/src/pages/devices/EditMatrixNamesField.js
```diff
@@ -0,0 +1,315 @@
...
+                set(newpersonalNames, `${ioKey}.${source}.name`, '');
+            }
+            if (isEmpty(get(newpersonalNames, `${ioKey}.${source}.channels`))) {
+                delete newpersonalNames[source];
+            }
+            return newpersonalNames;
+        });
+    };
+
+    const saveOverrideName = () => {
```
orfar1994, line 67: `saveOverrideName ` to `savePersonalName `

Development/src/pages/devices/EditMatrixNamesField.js
```diff
@@ -0,0 +1,315 @@
...
+            } else {
+                set(newpersonalNames, `${ioKey}.${source}.name`, value);
+            }
+            return newpersonalNames;
+        });
+        setDisplayRenameField(false);
+        setDisplayEditIcon(false);
+    };
+
+    const cancleOverrideName = () => {
```
orfar1994, line 88: `cancleOverrideName ` to `cancleEditName `

garethsb-sony:
Thanks for opening the PR, @orfar1994. I have had a play with this. The logic of both "personal" names and filtering is basically fine, but I have some concerns about the user interface.

*Rename UI*

* I don't like the way the UI resizes as you simply move the mouse over elements (hover, without clicking). At one time I got it into a state where hovering over a name caused it to show the edit button which resized and meant the mouse was no longer over it, so the UI flickered back and forth.
* I am concerned that the edit fields can be too small when the column is narrow, only two or three characters visible.
* I am not sure about being able to edit multiple names at the same time
* I think the presence of ✖️ (cancel) and the bin (delete) icon is a little confusing (delete should be presented at same time as edit)

My suggestion is that the rename functionality should move into the (now interactive) tooltip for the cell. The "personal" name in the tool tip can have edit and delete icons next to it and the edit field can be a sensible consistent width.

(I am also not sure about the word "personal", but this can be easily changed later obviously. Maybe "local", "user", "custom",... Brainstorm required!)

orfar1994:
so the "personal" title (with the edit & delete icon) in the tooltip will be visible always even if we don't have personal name ?

garethsb-sony:
Following on...

*Filtering*

* I think it would be simpler for the user if one filter searched both the "personal" and "original" labels?
* I'd like to add ID filters as well for consistency with other pages (I've just added ID to the tool tips on master)
* Can we group Output and Input filters in the drop down, with a divider between? Or have separate Add Input Filters and Add Output Filters?
* Limiting label length is not a filter, it should be in a separate UI somehow, perhaps under a settings (cog) menu? Or at least separated by a divider for now? (Very minor point but it also needs a minimum of at least one character.)
* A settings menu would also be a place to add a Delete All Personal Names function?
* We need to think about the Routable Inputs filter.
  * At the moment if you select a particular Input it shows Outputs that mention that Input but not Outputs that have no routing constraints. If the user is interested in whether they can route a particular Input, those should perhaps be included? 
  * Is it searching personal and original names?
  * I wonder if we should make Routable Inputs a Boolean or Const Filter that filters the Outputs based on the current filtered list of Inputs? Although the name seems a bit backwards then, and suggests we ought also have the opposite function of filtering the Inputs based on which can be routed to the current selection of Outputs...

More to come...

garethsb-sony:
> so the "personal" title (with the edit & delete icon) in the tooltip will be visible always even if we don't have personal name ?

I wasn't completely sure, what do you think? Probably needs prototyping to see.

garethsb-sony:
One issue with using tooltips a lot in this page is how the app performs on touchscreen (tablet). The whole app is already not perfect on small screen of course, but it is something to consider...

garethsb-sony:
*Rename UI*

* I also remember I was quite confused initially by the fact that the rows automatically re-sort as soon as you ✅ the edit field. It makes me wonder whether alphabetical sorting on label is a bad idea, and instead we want users to be able to move rows and columns manually? Either by drag handles or (on Inputs) up/down buttons that swap the row with the previous/next row... I think the initial sort should be by ID then.
* Of course, editing the label can mean the current filter will now exclude the new value, so we can't avoid that...

Development/src/pages/devices/EditMatrixNamesField.js
garethsb-sony: Cancel not cancle 👍

orfar1994:
image


image

garethsb-sony:
What do you think? I think that could work well. We can discuss on today's call.

orfar1994:
> _Rename UI_
> 
> * I also remember I was quite confused initially by the fact that the rows automatically re-sort as soon as you ✅ the edit field. It makes me wonder whether alphabetical sorting on label is a bad idea, and instead we want users to be able to move rows and columns manually? Either by drag handles or (on Inputs) up/down buttons that swap the row with the previous/next row... I think the initial sort should be by ID then.
> * Of course, editing the label can mean the current filter will now exclude the new value, so we can't avoid that...

actually I did it in purpose 

> What do you think? I think that could work well. We can discuss on today's call.

I agree. it looks better now. 
I think the icon should be in the same line of the "Name".
if personal name exists, than the original name title will appear and the personal name will appear under the "name" title.

orfar1994:
I agree. it looks better now. 
I think the icon should be in the same line of the "Name".
if personal name exists, than the original name title will appear and the personal name will appear under the "name" title.

garethsb-sony:
OK, that could work...

I assume the delete (bin) icon would be hidden if there isn't currently a user-provided name?
And that we should add tooltips for channels too of course?

We certainly need to be consistent in how we refer to the two kinds of names/labels.

Suggestions instead of "personal":

* user name/label (nope, "user name" is bad!)
* custom name/label
* local name/label
* display name/label
* override name/label (too techy)

Alternatively, if we use "Name" to mean user-provided-if-present-otherwise-underlying-name, and have a field for the underlying name that appears/disappears depending on whether there is a user-provided name, suggestions for that underlying name field:

* original name/label
* device name/label
* API name/label

None of these are great...

garethsb-sony:
Also, pressing Return in the edit field should choose the ✔️ so you don't have to click it explicitly to set the new value.

garethsb-sony:
Hi Or, thanks for all the work on this.

> At the moment the key is only based on the inputId/outputId? Did it ought to include the Device ID or something, since inputId and outputId are only unique locally within the Channel Mapping API instance?

Are you OK to resolve this issue?

orfar1994:
commit 19773b5 contain the fixes for - 

- tooltip jumps right after pressing on edit.
- automatically focus to the new edit box
- pressing on 'enter' save the new name.
- saving the new names under "device id" 
- when the new name is equal to the API name -> delete new name. 
- while editing (until you press tick or cross) the tooltip would not vanish even if the mouse cursor moves away and prevent any other tooltip appearing

Development/src/pages/devices/ChannelMappingMatrix.js
```diff
@@ -48,129 +62,321 @@ const TooltipDivider = withStyles({
...
+    truncateValue,
+    tooltipOpen,
+}) =>
     outputs.map(([outputId, outputItem]) => (
         <StyledTableCell
             align="center"
-            colSpan={isExpanded(outputId) ? outputItem.channels.length : 1}
+            colSpan={
+                isExpanded(outputId)
+                    ? Object.keys(outputItem.channels).length
```
garethsb-sony, line 372: Why does this need to use `Object.keys`? And similarly in a few other places that now have `Object.keys` or `Object.entries`.

Development/src/pages/devices/ChannelMappingMatrix.js
```diff
@@ -48,129 +62,321 @@ const TooltipDivider = withStyles({
     },
 })(Divider);
 
-const getOutputTooltip = (outputId, outputItem, io) => {
+const InteractiveTooltip = ({
+    getTooltip,
+    display,
+    tooltipOpen,
```
garethsb-sony, line 68: This means not "is open" but "stay open" if I understand correctly?
Just thinking about this after playing with the UI now this is implemented, maybe we actually just want to know whether an interactive element in the tooltip currently has the focus, and keep it open while that is the case. I.e. maybe clicking off a tooltip once you start editing should cancel the edit?
Need to think more!

Development/src/pages/devices/ChannelMappingMatrix.js
```diff
@@ -48,129 +62,321 @@ const TooltipDivider = withStyles({
     },
 })(Divider);
 
-const getOutputTooltip = (outputId, outputItem, io) => {
+const InteractiveTooltip = ({
+    getTooltip,
+    display,
```
garethsb-sony, line 67: Rather than a `display` function, I think the idiomatic React way would be to take the children and pass them on to the underlying `Tooltip`.

So user would write:
```
<InteractiveTooltip ...>
    <SomeComponent />
    <AnotherComponent />
</InteractiveTooltip>
```

And `InteractiveTooltip` would use [`props.children`](https://codeburst.io/a-quick-intro-to-reacts-props-children-cb3d2fce4891).

Development/src/pages/devices/ChannelMappingMatrix.js
```diff
@@ -48,129 +62,321 @@ const TooltipDivider = withStyles({
...
-            <Typography variant="body2">
-                {get(inputItem, `caps.reordering`) ? (
-                    <DoneIcon />
-                ) : (
-                    <ClearIcon />
-                )}
-            </Typography>
-        </>
-    );
+const getOutputTooltip = (
```
garethsb-sony, line 101: Similar to comment about `display` function in `InteractiveTooltip`, I think idiomatically this and the other `get...Tooltip` functions  would be a React component called e.g. `OutputTooltipContent`.

garethsb-sony (commented):
🚧 Code review in progress 🚧

I'm really happy to have this extra functionality, thank you, @orfar1994. I have not finished reviewing this - at the moment, I have added a lot of comments that are questions or unfinished thoughts rather than definite suggestions for changes!

Development/src/pages/devices/ChannelMappingMatrix.js
```diff
@@ -36,12 +37,55 @@ import { useJSONSetting } from '../../settings';
...
+
+// Midline Horizontal Ellipsis for when columns have been collapsed
+const HorizontalEllipsisButton = props => (
+    <IconButton size="small" children={'\u22ef'} {...props} />
+);
+
+// Vertical Ellipsis for when rows have been collapsed
+const VerticalEllipsisButton = props => (
+    <IconButton size="small" disabled children={'\u22ee'} {...props} />
+);
```
garethsb-sony, line 83: We discussed that `⋮` may look too much like a menu icon...

(https://twitter.com/lukew/status/591296890030915585)

The checked and unchecked states of disabled `MappingButton`s are also quite hard to distinguish, so we may want to revisit the matrix presentation after user feedback.

Development/src/pages/devices/ChannelMappingMatrix.js
```diff
@@ -48,129 +62,321 @@ const TooltipDivider = withStyles({
...
+            basePath="/flows"
+            label="Flows"
+            source="source_id"
+            reference="flows"
+            target="source_id"
+            link="show"
+            style={{
+                margin: 1,
+                padding: 1,
+            }}
```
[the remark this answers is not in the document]
garethsb-sony: I had a brief play with the `style` attributes earlier, and I'm still a bit confused. I think the real problem is that we are using react-admin's `ChipField`, `TextField`, etc. outside of a `Show` view, which means some of the structural layout and styling is different than expected - for example normally the `SimpleShowLayout` would automatically create the `InputLabel` component for each field based on the `label` prop, whereas here we're doing that manually with simple inline text. We may actually want to reuse `SimpleShowLayout` or create a `TooltipLayout`, injecting the `outputItem` as the `record`, etc. and similarly for the other tooltips.

Development/src/pages/devices/ChannelMappingMatrix.js
garethsb-sony: I hope this behaviour will be achievable by [detecting focus entering and leaving the component tree] (https://reactjs.org/docs/events.html#detecting-focus-entering-and-leaving).

Development/src/pages/devices/ChannelMappingMatrix.js
orfar1994: I convert the channels list to Object.
I save the new names of the channels by the index and if I filter the list , the index change and than the name is not correct...

Development/src/pages/devices/ChannelMappingMatrix.js
orfar1994: lets talk, I don't understand exactly what do you mean :)

Development/src/pages/devices/ChannelMappingMatrix.js
```diff
@@ -994,27 +994,40 @@ const sortByIOName = (ioObject, getCustomName) => {
 };
 
 const ChannelMappingMatrix = ({ record, isShow, mapping, handleMap }) => {
-    const [expandedCols, setExpandedCols] = useState([]);
-    const [expandedRows, setExpandedRows] = useState([]);
+    const [collapsedState, setCollapsedState] = useJSONSetting(
```
garethsb-sony, line 997: This is a good idea, except that it also persists when the user leaves this Device and is applied when the user goes to another Device's Active Map. Unlike the custom names/labels, I don't think the collapse state should be maintained per-Device, it should be reset when leaving the Device. Do you agree?

orfar1994: agree but have no idea how to do it...

Development/src/pages/devices/ChannelMappingMatrix.js
garethsb-sony: Ah, makes sense, thank you!

Development/src/pages/devices/ChannelMappingMatrix.js
garethsb-sony: OK, no problem.

Development/src/pages/devices/EditMatrixNamesField.js
```diff
@@ -310,7 +302,7 @@ export const EditableChannelLabelField = ({
                     <Typography variant="body2" display="inline">
                         {getChannelLabel()}
                     </Typography>
-                    {displayEditIcon && (
+                    {!displayEditTextField && (
```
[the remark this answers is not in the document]
orfar1994: fixed it :)

Development/src/pages/devices/EditMatrixNamesField.js
```diff
@@ -135,7 +131,7 @@ export const EditableIONameField = ({
                     <Typography variant="body2" display="inline">
                         {getName()}
                     </Typography>
-                    {displayEditIcon && (
+                    {!displayEditTextField && (
```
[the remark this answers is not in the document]
orfar1994: fixed it :)

Development/src/pages/devices/ChannelMappingMatrix.js
orfar1994: there is easy way to do it when leaving the active map tab.
```
 <Route
      exact path={`${props.basePath}/${props.id}/show/active_map`}
      onLeave={ removeItem }
    >
```

orfar1994: what do you think ?

garethsb-sony: sounds reasonable, I think

Development/src/pages/devices/ChannelMappingMatrix.js
orfar1994: I added commit that handling with this issue. 
when the user click away the tooltip (on edit mode) , the tooltip disappeared.

garethsb-sony:
> Thanks. this is better. However, I will still look for a way to improve this, because switching tabs causes a re-render of the show view, which means that the state is removed when switching between the tabs, not just when leaving the Device page.

Resolved by adding the empty dependencies array to the `useEffect` hook.

Development/src/pages/devices/ChannelMappingMatrix.js
garethsb-sony: @rhastie, I pushed a commit to fade-back the unchecked icons and ellipses. I think this simple change improves the UX considerably!

rhastie: Yes - That is looking better... Is the "ellipse-tick icon" at 50% transparency (probably wrong term!)? Could we also slightly fade-forward the "ellipse-tick icon" a little bit so it was half-way between 0 and 100% transparency? This means we can then still use the much darker 0% transparency for the "staged map" screen to show an "active change" etc...

garethsb-sony: The empty circle and the three-dots icons are now all faded back by applying 0.3x opacity. On top of that, all the icons in the Active Map, including those ones _and_ the ticked circle, are flagged as 'disabled' for which Material UI seems to apply 0.26x opacity by default. We want the 'disabled' state in the Active Map view to be significantly different to the 'enabled' state in the Staged Map, but 0.4x might be a compromise...

rhastie: Sounds a plan... IMHO we can dial the opacity back up if needed when we get to the Staged Map

garethsb-sony:
> When I have a bit more time, I want to look at refactoring `EditableIONameField` and `EditableChannelLabelField` into a single React element. It seems like maybe separate `deviceID`, `ioKey`, `source`, etc. are perhaps not necessary and a single argument that is e.g. set to `${deviceID}.${ioKey}.${source}` in the case of Input/Output Name could be passed in.

I've had a go at this over the course of several commits.

Development/src/pages/devices/EditMatrixNamesField.js
```diff
@@ -0,0 +1,335 @@
...
+import ClearIcon from '@material-ui/icons/Clear';
+import DoneIcon from '@material-ui/icons/Done';
+import { get, has, isEmpty, set } from 'lodash';
+
+export const EditableIONameField = ({
+    defaultValue,
+    source,
+    label,
+    customNames,
+    setCustomNames,
```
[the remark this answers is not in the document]
garethsb-sony: I think the idiomatic approach to reading and writing the custom names would be to use the [`useContext` hook](https://reactjs.org/docs/hooks-reference.html#usecontext). After a quick play, I think this would save 100+ lines of passing things through the nested components to where they are used. Something to look at later...

Development/src/pages/devices/EditMatrixNamesField.js
```diff
@@ -0,0 +1,335 @@
...
+                <>
+                    <TextField
+                        label={label}
+                        variant="filled"
+                        margin="dense"
+                        value={value}
+                        onChange={event => setValue(event.target.value)}
+                        inputRef={inputRef}
+                        fullWidth={true}
+                        onKeyPress={event => {
```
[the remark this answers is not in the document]
garethsb-sony: Seems to be no issue.

Development/src/pages/devices/EditMatrixNamesField.js
```diff
@@ -0,0 +1,335 @@
...
+                            if (event.key === 'Enter') {
+                                saveCustomName();
```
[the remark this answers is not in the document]
garethsb-sony: Seems to be no issue.

Development/src/pages/devices/EditMatrixNamesField.js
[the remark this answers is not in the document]
garethsb-sony: Both the custom names state and the tooltip modal state are now managed using context, provided by the ChannelMappingMatrix component itself and consumed only where necessary.

Development/src/pages/devices/FilterMatrix.js
```diff
@@ -0,0 +1,326 @@
+import get from 'lodash/get';
+import has from 'lodash/has';
+import { getCustomChannelLabel, getCustomName } from './ChannelMappingMatrix';
+
+const conditionGroup = (filterGroup, ...theArgs) =>
```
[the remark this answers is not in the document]
garethsb-sony: This code has now gone, but for what it's worth, I think the idiomatic JavaScript would probably use `Array.some` or `Array.every` or the Lodash equivalents.

Development/src/pages/devices/FilterMatrix.js
```diff
@@ -0,0 +1,326 @@
...
+    theArgs.reduce((previous, current) => {
+        if (filterGroup === 'or') {
+            return previous || current;
+        } else {
+            return previous && current;
+        }
+    });
+
+const channelIncludes = (label, channelLabelReg) => {
```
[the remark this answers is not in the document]
garethsb-sony: Still possibly opportunities to improve the use of `RegExp` but seems good enough for now.

Development/src/pages/devices/FilterMatrix.js
```diff
@@ -0,0 +1,326 @@
...
+    } else {
+        return (
+            blockSizeReg === undefined ||
+            isNaN(blockSizeReg) ||
+            item.caps.block_size === blockSizeReg
+        );
+    }
+};
+
+const filterReordering = (reorderingReg, item, filterGroup) => {
```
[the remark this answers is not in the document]
garethsb-sony: Fixed.

Development/src/pages/devices/FilterMatrix.js
```diff
@@ -0,0 +1,326 @@
...
+    return label.match(RegExp(`${channelLabelReg}`, 'i'));
+};
+
+const filterChannelLabel = (
+    channelLabelReg,
+    item,
+    customChannelLabel,
+    filterGroup
+) => {
+    if (filterGroup === 'or' && channelLabelReg === undefined) {
```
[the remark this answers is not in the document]
garethsb-sony: Fixed.

Development/src/pages/devices/ChannelMappingMatrix.js
garethsb-sony: Renamed as `tooltipModal` to indicate that it means an interactive tooltip is currently modal.

garethsb-sony: FWIW, `ClickAwayListener` now moved inside the tooltip content which means we don't get tens or hundreds of them firing!

garethsb-sony:
Thank you, @orfar1994, @rhastie. This is a great addition!

PR #98, opened 2025-08-05
Bump library versions for compatibility with Node20

jonathan-r-thorpe:
- Updated libraries to point of getting node-js working with Node20.
- Many libraries still deprecated, and so further upgrade work needed.


lo-simon (approved):
LGTM

PR #129, opened 2026-03-05
MXL Support

jonathan-r-thorpe:
- Prototype used to demonstrate MXL integration with NMOS


garethsb:
Domain before flow? 🤔

garethsb (approved):
👏

Development/src/components/mxlTransportParams.js
```diff
@@ -0,0 +1,164 @@
...
+    return legs.map(leg => {
+        const preparedLeg = { ...leg };
+        MXL_ID_PARAMS.forEach(param => {
+            if (!Object.prototype.hasOwnProperty.call(preparedLeg, param)) {
+                return;
+            }
+            const value = preparedLeg[param];
+            if (activating && isEmptyMxlValue(value)) {
+                preparedLeg[param] = 'auto';
+            } else if (value === '') {
```
lo-simon:
```suggestion
            if (value === '') {
```

lo-simon: For the MXL /staged endpoint, the empty string for `mxl_domain_id` and `mxl_flow_id` should not be converted to "auto" and should be set to null. BTW `auto` is invalid for the MXL receiver `mxl_flow_id` /staged endpoint.

The function `isEmptyMxlValue` is no longer in use and can be removed.

lo-simon (changes requested):
1. For the MXL /staged endpoint, the empty string for `mxl_domain_id` and `mxl_flow_id` should not be converted to "auto" and should be set to null.

2. The function `isEmptyMxlValue` is no longer in use and can be removed.

3. Should the `Transport File` edit box in the "EDIT MXL STAGED" tab be hidden?

Development/src/components/makeConnection.js
```diff
@@ -140,6 +141,14 @@ const makePatchDataWithTransportParams = (data, options) => {
             break;
         case 'urn:x-nmos:transport:websocket':
             break;
+        case 'urn:x-nmos:transport:mxl':
+            // Receivers must not use "auto" for mxl_flow_id (BCP-007-03)
+            legMap.forEach((senderLeg, receiverLeg) => {
+                if (get(patchParams[receiverLeg], 'mxl_flow_id') === 'auto') {
```
garethsb: Why would it have been set to `auto` ? `copyTransportParams`/`oneToOneTransportParams` should mean it's copied from the Sender's `/active` params which shouldn't contain `auto` unless something's gone wrong. And if it has, maybe we want to just attempt a "bad patch" and get an error, not silently adjust to `null`?

Development/src/pages/receivers/ReceiversShow.js
```diff
@@ -258,7 +258,10 @@ const ShowActiveTab = ({ record, ...props }) => {
                 >
                     <ReceiverTransportParamsCardsGrid record={record} />
                 </ArrayField>
-                <TransportFileViewer endpoint="$active.transport_file.data" />
+                {get(record, '$transporttype') !==
+                    'urn:x-nmos:transport:mxl' && (
+                    <TransportFileViewer endpoint="$active.transport_file.data" />
+                )}
```
garethsb: Probably a `null` transport file renders fine in the `TransportFileViewer`, because we can already have `null` in RTP senders/receivers too? So maybe don't bother with the transport check here and below?

lo-simon (approved):

PR #145, opened 2026-05-12
Pin Dockerfile to node:22 LTS to restore Docker build

### LINKED ISSUE #144: Docker build fails: `yarn: not found` in `node:latest`
## Summary

Building the Docker image with the documented `Development/Dockerfile` fails on a clean machine. The first `RUN yarn` step exits 127 because the `yarn` binary is no longer present in the upstream `node:latest` image.

## Reproduce

```
git clone https://github.com/sony/nmos-js.git
cd nmos-js
docker build -t nmos-js Development
```

Output:

```
#10 [build 4/6] RUN yarn
#10 5.722 /bin/sh: 1: yarn: not found
#10 ERROR: process "/bin/sh -c yarn" did not complete successfully: exit code: 127
------
 > [build 4/6] RUN yarn:
5.722 /bin/sh: 1: yarn: not found
------
Dockerfile:4
--------------------
   2 |     WORKDIR /usr/src/app
   3 |     COPY package.json yarn.lock ./
   4 | >>> RUN yarn
   5 |     COPY . ./
   6 |     RUN yarn build
```

## Root cause

The Dockerfile is pinned to `FROM node:latest`. As of the current `node:latest` (Node 26.1.0, Debian Trixie) neither `yarn` nor `corepack` is installed in the image:

```
$ docker run --rm node:latest sh -c 'node --version; yarn --version 2>/dev/null || echo "yarn: missing"; corepack --version 2>/dev/null || echo "corepack: missing"'
v26.1.0
yarn: missing
corepack: missing
```

Older Node lines (`node:18`, `node:20`, `node:22`, `node:24`) still bundle `yarn 1.22.22` and `corepack`, so this regression is specific to the Node 26 series that `node:latest` now tracks.

## Why CI doesn't catch this

`.github/workflows/build-test.yml` uses `actions/setup-node@v4` on `ubuntu-22.04`, where `yarn` is provided by the GitHub-hosted runner image rather than by the Node Docker image. The Dockerfile build path is not exercised by CI, so the breakage is invisible in the matrix.

## Relationship to #83

This is distinct from #83 (the `ERR_PACKAGE_PATH_NOT_EXPORTED` postcss failure during `yarn build`). #83 was effectively addressed for the build step by #98 ("Bump library versions for compatibility with Node20"). The failure reported here is earlier in the build — `yarn` is missing before any code is compiled — and is a consequence of recent upstream changes to the official `node` images, not the project's dependency tree.

## Proposed fix

Pin `Development/Dockerfile` to the active LTS line — `FROM node:22 AS build` — which still ships `yarn 1.22.22` and aligns with the Node 20+ compatibility work in #98. PR to follow.

mattlamb99:
## Summary

- Pins `Development/Dockerfile` from `FROM node:latest` to `FROM node:22 AS build`.
- Restores the documented `docker build -t nmos-js Development` quick-start, which currently fails with `yarn: not found` because `node:latest` (Node 26.1.0) no longer bundles `yarn` or `corepack`.
- One-line change; no other modifications.

Fixes #144.

## Background

`node:latest` historically shipped a pre-installed `yarn` binary, which `RUN yarn` in the Dockerfile relied on. Recent upstream Node Docker images have dropped both `yarn` and `corepack`:

```
$ docker run --rm node:latest sh -c 'node -v; yarn -v 2>/dev/null || echo missing; corepack -v 2>/dev/null || echo missing'
v26.1.0
missing
missing
```

The active LTS line (`node:22`) still ships `yarn 1.22.22`, so pinning to it is the minimum change that restores the build. This is also consistent with the Node 20+ dependency work that landed in #98.

This regression is invisible in CI because `.github/workflows/build-test.yml` uses `actions/setup-node@v4` on a GitHub-hosted runner that provides `yarn` system-wide; the Dockerfile path is not exercised by the matrix.

## Test plan

- [x] `docker build -t nmos-js Development` succeeds on a clean machine with this change (verified locally; nginx-served React shell returns HTTP 200 from the resulting image).
- [ ] Reviewer runs `docker build -t nmos-js Development` against `master` to confirm the failure, then against this branch to confirm the fix.
- [ ] Optional: consider adding a Docker build job to `.github/workflows/build-test.yml` so future regressions in the Docker path are caught in CI. Not included here to keep the diff focused.

### DIFF
--- a/Development/Dockerfile
+++ b/Development/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:latest AS build
+FROM node:22 AS build
 WORKDIR /usr/src/app
 COPY package.json yarn.lock ./
 RUN yarn


jonathan-r-thorpe (approved):
LGTM

lo-simon (approved):
LGTM

PR #146, opened 2026-05-25
Nmos CVE

Ov3rIT:


jonathan-r-thorpe:
This looks like a local specialization of nmos-js rather than a bug fix or functional improvement. Perhaps this PR has been generated in error?

lo-simon:
Yes, I agree with @jonathan-r-thorpe and @garethsb that this is for local use rather than a bug fix.

PR #148, opened 2026-06-03
Add NMOS Connection API Bridge (Envoy) and client support

mattlamb99:
Provides browser access to IS-05 Connection APIs for Devices that are not reachable from the client, following the Connection API Bridge design proposed in review (thanks @garethsb).

This replaces the original server-side Node proxy in this PR entirely. The branch has been rebased; nginx is untouched and nothing in the existing deployment changes unless the bridge is used.

## NMOS Connection API Bridge (`ConnectionBridge/`)

- Envoy proxies `/x-nmos-bridge/v1.0/devices/{device_id}/connection/{version}/{sub-path}` to `{href}/{sub-path}`, where `href` comes only from registered Device `controls` matching `urn:x-nmos:control:sr-ctrl/{version}`. Arbitrary URLs cannot be proxied and the registry needs no changes.
- An adapter service polls the Query API and generates Envoy routes and clusters with the proposed deterministic naming, atomically replacing dynamic configuration files which Envoy reloads via filesystem watch. The adapter performs no proxying and no health checks.
- Multiple hrefs for a Device/version pair become prioritized failover candidates (private IPs, then private DNS, then others). Envoy health checks the candidates and fails over between priority levels. Separate Connection API versions are never merged.
- Methods restricted to GET, POST, PATCH, OPTIONS. GETs may be retried; POSTs and PATCHes are not. POST and PATCH are access logged. Request bodies limited to 1 MB. Unknown targets get 404 and disallowed methods 405, in NMOS error format.
- Phase 1 plus health checking and multi-endpoint failover from Phase 2 are implemented. Not yet: response size limits, HTTPS upstreams, authentication translation, mTLS, xDS control plane.

## Client

A new Connection Bridge setting (default off) with three modes:

- **No Bridge**: always use Device control hrefs directly (current behavior).
- **Auto Bridge**: the preferred access sequence. Try control hrefs directly, fall back to the bridge if they are inaccessible, cache the successful path per Device.
- **Forced Bridge**: always use the bridge.

One observation from testing prompted the third mode: in Auto, the first access to an unreachable Device has to wait for the direct attempt to fail (the existing 5 second timeout) before the bridge is tried. The per-Device cache removes that wait on subsequent accesses, but when it is known that no Device is reachable from the browser, Forced Bridge skips the direct attempt entirely.

POST and PATCH follow whichever access path was resolved for the Device and are never retried via alternate paths. The client expects the bridge on the same host as the configured Query API.

## Testing

Tested two ways. Against a stub registry and stub IS-05 Device: GET and PATCH to staged through the bridge, 405 for disallowed methods, 404 for unknown targets, CORS preflight, prioritized failover cluster generation for a Device with multiple hrefs, and hot reload of routes when a Device is added to the registry without restarting Envoy. Against a live registry and production devices: bridge GET responses byte-identical to direct Device access, a no-op PATCH to a real Device through the bridge (200, correctly routed to the registered control href, captured by the POST/PATCH access log), discovery generating bridge targets across all registered Devices and API versions, and the client in all three modes including fallback for Devices on networks the browser cannot reach.


mattlamb99:
A couple of points I'd like to get the maintainers' input on:

**Dockerfile (nginx to Node/Express):** Serving the built app and the proxy from a single process was the simplest way to get this working, but it means we are replacing nginx. I'm open to changing this approach. For example, we could keep nginx for the static assets and run the proxy as a separate small service, or we could make the proxy opt-in at deploy time. Let me know what fits best with your maintenance preferences.

**HTTPS nodes with private CAs:** The proxy uses Node's fetch, which automatically rejects certificates it can't validate. I avoided adding a TLS-verification bypass, so any nodes served over HTTPS with a private or self-signed CA won't be reachable through the proxy right now. I wanted to flag this as a known limitation for your deployments. If we want to support this, I'm happy to look into a clean way to trust custom CAs, perhaps via something like NODE_EXTRA_CA_CERTS.

garethsb:
This is how I'd approach this.

# NMOS Connection API Bridge

## Goal

Provide browser-accessible proxy access to AMWA IS-05 Connection APIs exposed by Devices registered in an NMOS Registry, where the browser may not have network access to the Device APIs directly.

## Public Bridge API

### Root Path

```text
/x-nmos-bridge/v1.0
```

### Device Connection API Mapping

```text
http://registry.example.com/x-nmos-bridge/v1.0/devices/{device_id}/connection/{version}/{sub-path}
```

Proxies to:

```text
{href}/{sub-path}
```

Where:

- `href` is taken from the Device resource `controls` entry matching:
  - `urn:x-nmos:control:sr-ctrl/{version}`
- `version` is the Connection API version being accessed.

### Example

Public:

```text
PATCH /x-nmos-bridge/v1.0/devices/{device_id}/connection/v1.1/single/receivers/{receiver_id}/staged
```

Upstream:

```text
PATCH http://device.example.local/x-nmos/connection/v1.1/single/receivers/{receiver_id}/staged
```

---

# Architecture

```text
Browser
    |
    v
 Envoy
    |
    +--> Registry APIs
    |
    +--> NMOS Connection API Bridge
             |
             v
        Device Connection APIs
```

Components:

- Registry
- Envoy
- Envoy Adapter
- Browser Application

---

# Registry Responsibilities

The registry remains the source of truth.

Requirements:

- Continue exposing standard NMOS APIs.
- Continue exposing Device resources.
- Continue exposing Device `controls` entries.
- No response rewriting required.
- No Envoy-specific behavior required.

---

# Envoy Responsibilities

Envoy provides HTTP proxying.

Responsibilities:

- Route registry API requests.
- Route bridge requests.
- Forward GET, POST and PATCH.
- Preserve request bodies and query strings.
- Enforce allowlists.
- Apply timeouts.
- Perform health checks.
- Perform failover where configured.

---

# Envoy Adapter Responsibilities

The adapter converts registry state into Envoy configuration.

Responsibilities:

- Discover Devices from the registry.
- Extract Connection API controls.
- Generate Envoy routes.
- Generate Envoy clusters.
- Write dynamic configuration files atomically.
- Regenerate configuration when registry state changes.

The adapter should not perform health checks itself.

---

# URL Mapping Rules

Given:

```json
{
  "id": "device-uuid",
  "controls": [
    {
      "type": "urn:x-nmos:control:sr-ctrl/v1.1",
      "href": "http://10.20.30.40:8080/x-nmos/connection/v1.1/"
    }
  ]
}
```

Generate:

```text
/x-nmos-bridge/v1.0/devices/device-uuid/connection/v1.1/{sub-path}
```

Proxy target:

```text
http://10.20.30.40:8080/x-nmos/connection/v1.1/{sub-path}
```

Rules:

- Preserve query strings.
- Preserve HTTP methods.
- Preserve request bodies.
- Remove bridge prefix.
- Append remaining path to the selected control href.
- Ensure exactly one slash between href and sub-path.

---

# Supported Methods

Allowed:

```text
GET
POST
PATCH
OPTIONS
```

Retry policy:

```text
GET      Retry allowed
POST     No automatic retry
PATCH    No automatic retry
```

---

# Multiple Advertised Connection API URLs

Devices may expose multiple usable Connection API alternatives.

The adapter should:

- Group alternatives by Device and API version.
- Generate endpoint priorities.
- Generate cluster failover configuration.

Envoy should:

- Determine health.
- Select healthy endpoints.
- Perform failover when required.

The adapter should not determine runtime health.

---

# Security Constraints

The bridge must not behave as an open proxy.

Requirements:

- Targets must originate from registered Device controls.
- Public requests use Device IDs only.
- Arbitrary URLs are forbidden.
- Only Connection API controls are eligible.
- Restrict methods to the allowlist.
- Apply request size limits.
- Apply response size limits.
- Strip hop-by-hop headers.
- Log POST and PATCH requests.

Initial deployment:

```text
Browser -> Envoy: HTTP
Envoy -> Device: HTTP
```

---

# Browser Application Behavior

Preferred access sequence:

1. Use Device control href directly.
2. If inaccessible, use bridge URL.
3. Cache successful access path per Device.

Bridge URL pattern:

```text
/x-nmos-bridge/v1.0/devices/{device_id}/connection/{version}/{sub-path}
```

POST and PATCH requests should not automatically retry via alternate paths.

---

# Adapter Algorithm

```text
loop:

  discover Devices

  for each Device:

      for each control:

          if control.type is Connection API:

              create mapping

  generate Envoy configuration

  validate

  atomically replace configuration files

  repeat
```

---

# Adapter Mapping Model

A single NMOS Device may advertise several Connection API versions, and each version may have one or more usable backend `href` alternatives.

The adapter should treat each unique combination of:

```text
Device ID
Connection API version
```

as a separate bridge target.

## Input Model

Example Device controls:

```json
{
  "id": "device-uuid",
  "controls": [
    {
      "type": "urn:x-nmos:control:sr-ctrl/v1.0",
      "href": "http://10.20.30.40:8080/x-nmos/connection/v1.0/"
    },
    {
      "type": "urn:x-nmos:control:sr-ctrl/v1.1",
      "href": "http://10.20.30.40:8080/x-nmos/connection/v1.1/"
    },
    {
      "type": "urn:x-nmos:control:sr-ctrl/v1.1",
      "href": "http://device-a.local:8080/x-nmos/connection/v1.1/"
    }
  ]
}
```

This produces two bridge targets:

```text
device-uuid / v1.0
device-uuid / v1.1
```

The `v1.1` target has two backend href candidates.

## Public Route Mapping

For each Device/version pair, generate one public bridge prefix:

```text
/x-nmos-bridge/v1.0/devices/{device_id}/connection/{connection_api_version}/
```

Examples:

```text
/x-nmos-bridge/v1.0/devices/device-uuid/connection/v1.0/
/x-nmos-bridge/v1.0/devices/device-uuid/connection/v1.1/
```

Each route proxies sub-paths to the selected backend href for that same Connection API version.

## Envoy Naming Convention

Generated names should be deterministic and safe for Envoy configuration.

Recommended logical names:

```text
route:
  nmos_bridge_device_{safe_device_id}_connection_{safe_connection_version}

cluster:
  nmos_bridge_device_{safe_device_id}_connection_{safe_connection_version}

endpoint:
  nmos_bridge_device_{safe_device_id}_connection_{safe_connection_version}_{index}
```

Where:

```text
safe_device_id:
  Device ID with characters outside [A-Za-z0-9_] replaced by _

safe_connection_version:
  Version with dots replaced by _, for example v1.1 -> v1_1
```

Example:

```text
Device ID:
  3a2c5f90-6c8a-4e0b-83a1-90f2cbbf53c1

Connection API version:
  v1.1

Cluster name:
  nmos_bridge_device_3a2c5f90_6c8a_4e0b_83a1_90f2cbbf53c1_connection_v1_1
```

## Backend Href Candidate Handling

For each Device/version pair, the adapter should collect all eligible `href` values.

Eligible controls must satisfy:

```text
type matches urn:x-nmos:control:sr-ctrl/{version}
href scheme is supported
href is syntactically valid
href path corresponds to the advertised Connection API version
```

Initial supported scheme:

```text
http
```

Later supported schemes may include:

```text
https
```

The adapter should preserve the full upstream base path from `href`.

Example:

```text
href:
  http://10.20.30.40:8080/x-nmos/connection/v1.1/

public:
  /x-nmos-bridge/v1.0/devices/device-uuid/connection/v1.1/single/senders

upstream:
  http://10.20.30.40:8080/x-nmos/connection/v1.1/single/senders
```

## Multiple Hrefs for the Same Device and Version

If multiple hrefs exist for the same Device/version pair, the adapter should generate a single logical bridge target with multiple upstream candidates.

Preferred behavior:

```text
- All candidates serve the same Connection API version.
- Envoy selects a healthy candidate.
- Envoy fails over if the preferred candidate becomes unhealthy.
```

The adapter may assign priorities based on policy, for example:

```text
Priority 0:
  private IP address hrefs

Priority 1:
  private DNS hrefs

Priority 2:
  other reachable hrefs
```

Example:

```text
Device:
  device-uuid

Version:
  v1.1

Public route:
  /x-nmos-bridge/v1.0/devices/device-uuid/connection/v1.1/

Backend candidates:
  priority 0: http://10.20.30.40:8080/x-nmos/connection/v1.1/
  priority 1: http://device-a.local:8080/x-nmos/connection/v1.1/
```

## Separate Versions Must Not Be Merged

Different Connection API versions must be mapped independently.

For example, these are separate bridge targets:

```text
/x-nmos-bridge/v1.0/devices/device-uuid/connection/v1.0/
/x-nmos-bridge/v1.0/devices/device-uuid/connection/v1.1/
```

The `v1.0` bridge route must not fail over to a `v1.1` backend href, and vice versa.

## Path Rewrite Rule

For a matched bridge route:

```text
/x-nmos-bridge/v1.0/devices/{device_id}/connection/{version}/{sub-path}
```

rewrite to:

```text
{selected_href}/{sub-path}
```

Rules:

```text
- The `{version}` path segment selects the backend candidate group.
- The `{sub-path}` is appended to the selected backend href.
- Query strings are preserved.
- The bridge API version `v1.0` is not the same as the Connection API version.
```

## Invalid or Ambiguous Inputs

The adapter should ignore, reject, or log controls that are not safe to proxy.

Examples:

```text
- Missing href
- Unsupported scheme
- Malformed URL
- Connection API type with invalid version syntax
- Duplicate hrefs after normalization
- href path that appears inconsistent with the advertised version
```

Duplicates should be de-duplicated before generating Envoy configuration.

## Mapping Output Example

Input controls:

```text
Device device-uuid:
  sr-ctrl/v1.0 -> http://10.20.30.40:8080/x-nmos/connection/v1.0/
  sr-ctrl/v1.1 -> http://10.20.30.40:8080/x-nmos/connection/v1.1/
  sr-ctrl/v1.1 -> http://device-a.local:8080/x-nmos/connection/v1.1/
```

Generated bridge mappings:

```text
/x-nmos-bridge/v1.0/devices/device-uuid/connection/v1.0/*
  -> http://10.20.30.40:8080/x-nmos/connection/v1.0/*

/x-nmos-bridge/v1.0/devices/device-uuid/connection/v1.1/*
  -> priority 0: http://10.20.30.40:8080/x-nmos/connection/v1.1/*
  -> priority 1: http://device-a.local:8080/x-nmos/connection/v1.1/*
```

---

# Implementation Phases

## Phase 1

- HTTP browser access
- HTTP upstream access
- File-based Envoy dynamic configuration
- GET, POST, PATCH support
- No response rewriting

## Phase 2

- Health checking
- Multi-endpoint failover
- Enhanced audit logging

## Phase 3

- HTTPS upstreams
- Authentication translation
- mTLS support
- Optional xDS control plane

---

# Non-Goals

Do not implement:

- Arbitrary URL proxying
- Registry response rewriting
- Custom forwarding APIs in the registry
- Automatic POST/PATCH retries
- Registry awareness of Envoy configuration formats

mattlamb99:
Thanks @garethsb, this design is much more thorough than what I had, and I have now implemented it. The branch has been reworked, the Node proxy and the nginx replacement are gone. And the PR description updated.

What is in place, following the proposal:

- Envoy does all the proxying. The public API is `/x-nmos-bridge/v1.0/devices/{device_id}/connection/{version}/{sub-path}`, targets are resolved exclusively from Device `controls` matching `urn:x-nmos:control:sr-ctrl/{version}`, and arbitrary URLs are rejected. The registry is unchanged.
- The adapter polls the Query API, generates routes and clusters with your naming convention, and atomically replaces file-based dynamic configuration which Envoy picks up via filesystem watch. Multiple hrefs per Device/version become prioritized candidates (private IP, then private DNS, then other) and Envoy does the health checking and failover. Versions are never merged.
- GET/POST/PATCH/OPTIONS only, retries on GET only, POST and PATCH access logged, 1 MB request limit, NMOS-format 404/405 for unknown targets and disallowed methods.
- The client implements the preferred access sequence (although configurable): direct href first, bridge on failure, successful path cached per Device, and no POST/PATCH retry across paths.

One thing I noticed while testing against a live registry (devices on routable and non-routable networks): with the automatic sequence, the first access to an unreachable Device has to wait for the direct attempt to fail (the existing 5 second timeout) before the bridge is tried. The per-Device cache hides this after the first access, but it is noticeable cold. I have therefore made the client setting a three-way choice: No Bridge (default), Auto Bridge (the preferred sequence), and Forced Bridge (always use the bridge, for deployments where it is known the browser cannot reach any Device). 
Happy to drop Forced if you would rather keep the behavior purely automatic.

Phase 1 is implemented, plus health checking and multi-endpoint failover from Phase 2. Response size limits, HTTPS upstreams, authentication translation, mTLS and xDS remain future work.

garethsb:
@mattlamb99 Sorry, I haven't had a chance to try this out yet but the implementation looks pretty much like I imagined! One thing: I wonder whether using a Query API WebSocket subscription for Devices would be more elegant than polling the HTTP API? Would need to create a non-persistent subscription and handle the Device added/modified/removed events. Small bit of work to manage reconnecting to the WebSocket if the connection got broken, refreshing all mappings in that case.

mattlamb99:
@garethsb  Yeah - more elegant. However the map could silently drift if we ever missed a removed event or the socket half-opens without firing onclose (I don't think IS-04 has a heartbeat to detect it), would you want a low-frequency GET /devices poll (~5 min) kept as a self-healing backstop? Or do you'd rather stay WebSocket-only and rely on a ping/pong heartbeat for liveness, and lose the current full-refresh self-healingness?

garethsb:
I don't remember if the nmos-cpp WebSocket implementation supports ping pong, but that's worth checking.

I can't think of a reason you'd miss a remove (or any event) without an obvious error occurring?

However if there's a case, healing by closing and re-opening the WebSocket would work?

mattlamb99:
Okay, It doesn't look like nmos-cpp  supports ws ping pong. 
So plain WebSocket it is, no HTTP poll,  just relies on the full sync on ws (re)connect if there are any connection issues. 
Works well in my environment!

mattlamb99:
@garethsb  Have you had any chance to test this out? Has been working well for me :)

mattlamb99:
Merged both of your PRs, thanks! I made one small addition: REGISTRY_QUERY_WS_URL now applies the scheme default port when the override omits one, since the URL host setter otherwise leaves the advertised port in place. And updated the deployment docs to match. 
The branch is now rebased. Note for easy-nmos#37- the pinned commit is no longer on the branch; so that will need updating.

garethsb (approved):
LGTM. Thanks, @mattlamb99!

@lo-simon, @jonathan-r-thorpe, I recommend merging. The new bridge is opt-in.

garethsb:
Congrats, @mattlamb99! I updated https://github.com/rhastie/easy-nmos/pull/37, try out/review appreciated.

PR #152, opened 2026-06-17
Add MXL Controller Tests

jonathan-r-thorpe:
- Added automated BCP0070302 tests 
- Refactored TestingFacade:
  - TestingFacade no longer needs to be torn down between test suite runs
  - Instead the NMOS Testing tool signals which test suite is being run, and the TestingFacade configures itself accordingly
- Temporarily using branch of NMOS Testing that includes the BCP0070302 test suite (will be rolled back before merge)
- Updated README.md to reflect MXL support


lo-simon (approved):
LGTM

PR #157, opened 2026-07-02
Integrate IS-12 Browser into nmos-js

jonathan-r-thorpe:
- IS-12 Browser can be launched with a URI as a parameter
- From the Devices page, Control Protocol URIs can launch the IS-12 client which automatically connects to the endpoint

### DIFF
--- a/Development/src/components/ParameterRegisters.js
+++ b/Development/src/components/ParameterRegisters.js
@@ -103,6 +103,11 @@ export const CONTROL_TYPES = {
         label: 'Channel Mapping API',
         versions: ['v1.0'],
     },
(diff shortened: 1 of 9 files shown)


cristian-recoseanu:
I have tried the functionality locally against our implementation.
Everything seems to work as advertised as far as I can tell. Great effort @jonathan-r-thorpe.
We should try and expedite its review and merge as it would be extremely useful to the vendors in the Control & monitoring pioneers event.

lo-simon (approved):
LGTM

For future improvement, maybe we can launch the IS-12 browser/client if it is not already running.

PR #158, opened 2026-07-02
Add Status Monitor Panels to IS-12 Client

jonathan-r-thorpe:
- Visualizes the Sender and Receiver monitor panels status information
- A more compact representation than the raw object

Collapsed monitor blocks give status overview of contained monitors in header bar:
image
Expanded blocks show overall status of contained monitors:
image
Individual monitor panel can be expanded to see domain statuses:
image
Settings and methods panels can be expanded within monitor:
image

### DIFF
--- a/is12-client/src/client/EditablePropertyControl.jsx
+++ b/is12-client/src/client/EditablePropertyControl.jsx
@@ -27,7 +27,7 @@ const NUMERIC_TYPE_NAMES = new Set([
 
 function isValidSubmitValue(valueHolder, newValue) {
     if (valueHolder.name === 'userLabel') {
-        return newValue !== '' && newValue.length <= 30;
+        return newValue !== '' && (newValue?.length ?? 0) <= 30;
     }
 
     if (NUMERIC_TYPE_NAMES.has(valueHolder.datatype?.typeName)) {
--- a/is12-client/src/client/EditablePropertyControl.jsx
+++ b/is12-client/src/client/EditablePropertyControl.jsx
@@ -31,7 +31,7 @@ function isValidSubmitValue(valueHolder, newValue) {
     }
 
     if (NUMERIC_TYPE_NAMES.has(valueHolder.datatype?.typeName)) {
-        return newValue !== '' && !Number.isNaN(Number(newValue));
+        return newValue != null && newValue !== '' && !Number.isNaN(Number(newValue));
     }
 
     return newValue !== '';
@@ -75,8 +75,8 @@ export default function EditablePropertyControl({
     parameterName,
     compact = false,
 }) {
-    const datatype = valueHolder.datatype;
-    const value = valueHolder.value;
+    const datatype = valueHolder?.datatype;
+    const value = valueHolder?.value;
     const invariantType = undefined;
     const isNumericType = NUMERIC_TYPE_NAMES.has(datatype?.typeName);
 
@@ -85,6 +85,10 @@ export default function EditablePropertyControl({
     const [newInvariantType, setInvariantType] = useState(invariantType);
 
     const DataProvider = useContext(DataProviderContext);
+
+    if (!valueHolder) {
+        return null;
+    }
 
     const inputWidthSx = compact
         ? {
@@ -102,7 +106,8 @@ export default function EditablePropertyControl({
                     id="demo-simple-select"
                     sx={inputWidthSx}
                     size="small"
-                    value={value}
+                    displayEmpty
+                    value={value ?? ''}
                     onChange={(event) => {
                         setNewValue(event.target.value);
                         submitPropertyValue(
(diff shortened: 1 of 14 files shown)


is12-client/src/client/EditablePropertyControl.jsx
```diff
@@ -0,0 +1,289 @@
...
+    'NcUint16',
+    'NcUint32',
+    'NcUint64',
+    'NcFloat32',
+    'NcFloat64',
+]);
+
+function isValidSubmitValue(valueHolder, newValue) {
+    if (valueHolder.name === 'userLabel') {
+        return newValue !== '' && newValue.length <= 30;
```
cristian-recoseanu: nmos-js seems to be crashing against our IS-12 implementation at this line.
I think it's because it assumes userLabels can't be null?

Which isn't the case according to MS-05-02:
`[element("1p6")]                attribute    NcString?    userLabel;    // Scribble strip`

Something like this fixed the crash for me:
```suggestion
        return newValue !== '' && (newValue?.length ?? 0) <= 30;
```

jonathan-r-thorpe: Good spot!

cristian-recoseanu:
This is phenomenal work.
We should try and expedite its review and merge as it would be extremely useful to the vendors in the Control & monitoring pioneers event.

lo-simon (approved):
What great work to instruct a status-monitor panel. 
LGTM

**The `index`/`level` is not constrained to a minimum of 1 for the method control. I will try using Copilot to provide a suggestion to address it.**

is12-client/src/client/EditablePropertyControl.jsx
```diff
@@ -0,0 +1,294 @@
...
+    if (valueHolder.name === 'userLabel') {
+        return newValue !== '' && (newValue?.length ?? 0) <= 30;
+    }
+
+    if (NUMERIC_TYPE_NAMES.has(valueHolder.datatype?.typeName)) {
+        return newValue != null && newValue !== '' && !Number.isNaN(Number(newValue));
+    }
+
+    return newValue !== '';
+}
```
lo-simon:
```suggestion
function isValidSubmitValue(valueHolder, newValue, isMethod = false) {
    if (valueHolder.name === 'userLabel') {
        return newValue !== '' && (newValue?.length ?? 0) <= 30;
    }

    if (NUMERIC_TYPE_NAMES.has(valueHolder.datatype?.typeName)) {
        const numericValue = Number(newValue);

        if (newValue == null || newValue === '' || Number.isNaN(numericValue)) {
            return false;
        }

        if (isMethod && (valueHolder.name === 'index' || valueHolder.name === 'level')) {
            return numericValue >= 1;
        }

        return true;
    }

    return newValue !== '';
}
```

is12-client/src/client/EditablePropertyControl.jsx
```diff
@@ -0,0 +1,294 @@
...
+                    checked={Boolean(value)}
+                    onChange={handleChange}
+                    sx={compact ? { p: 0.25, ml: -0.25 } : undefined}
+                    inputProps={{ 'aria-label': valueHolder.name }}
+                />
+            </Box>
+        );
+    }
+
+    const canSubmit = isValidSubmitValue(valueHolder, newValue);
```
lo-simon:
```suggestion
    const canSubmit = isValidSubmitValue(valueHolder, newValue, isMethod);
    const hasMinimumOneConstraint =
        isMethod && isNumericType && (valueHolder.name === 'index' || valueHolder.name === 'level');

```

is12-client/src/client/EditablePropertyControl.jsx
```diff
@@ -0,0 +1,294 @@
...
+                                    index,
+                                    parameterName,
+                                    newInvariantType
+                                );
+                            }
+                        }}
+                        onChange={(event) => {
+                            setNewValue(event.target.value);
+                        }}
+                        value={newValue ?? ''}
```
lo-simon, line 229:
```suggestion
                        value={newValue ?? ''}
                        inputProps={hasMinimumOneConstraint ? { min: 1 } : undefined}                        
```

is12-client/src/client/EditablePropertyControl.jsx
lo-simon: Enforce that the `index`/`level` is constrained to a minimum of 1.

is12-client/src/client/EditablePropertyControl.jsx
lo-simon: Enforce that the `index`/`level` is constrained to a minimum of 1.

is12-client/src/client/EditablePropertyControl.jsx
lo-simon: Enforce that the `index`/`level` is constrained to a minimum of 1.

lo-simon (changes requested):
Enforce that the `index`/`level` is constrained to a minimum of 1. This is suggested by Copilot.

is12-client/src/client/EditablePropertyControl.jsx
jonathan-r-thorpe: Good catch - I think this bug was present before this PR

PR #160, opened 2026-07-15
Make IS-12 client LAF consistent with nmos-js

jonathan-r-thorpe:
- Updated look and feel of the IS-12 client to make it consistent with nmos-js
- Pass Node label to the IS-12 Client to use in title bar, and browser tab
image


lo-simon (approved):
LGTM

PR #162, opened 2026-07-20
Fix GetTransmissionErrorCounters not being rendered

cristian-recoseanu:
See https://specs.amwa.tv/nmos-control-feature-sets/branches/main/monitoring/#ncsendermonitor


lo-simon (approved):
LGTM

PR #163, opened 2026-07-25
Always include a requested activation in the connection PATCH

### LINKED ISSUE #97: No "activation" field in PATCH json when activating from Receivers->Staged tab or in "Connect" tab
When working with Receivers, there is no "activation" object present in the json sent to the target receiver node when attempting to activate through either the "Activation" button in the "Connect" tab (at the route http://localhost:3000/#/receivers/43f6e986-47cb-5181-992a-b2e332edaf00/show/connect) or in the "Staged/Edit" screen (at the route http://localhost:3000/#/receivers/43f6e986-47cb-5181-992a-b2e332edaf00). 

I would expect in either of those cases, the object in the PATCH request to the node would include an `activation` object with `"mode": "activate_immediate"`, but this is all I see in the patch body (data has been blanked for brevity, but did include SDP text):

```json
{
  "transport_params": [{ "rtp_enabled": true }],
  "sender_id": "3aa00e22-c761-5d04-ba87-d0d80727955b",
  "transport_file": {
    "data": "",
    "type": "application/sdp"
  }
}
```

I may be incorrect in my expectation though. I would think it _should_ include the `activation` object any time we want the staged params to take effect on the target resource and become the new `active` settings.

ZayanKhan-12:
Fixes #97

## Root cause

Exactly as suspected in the issue discussion: the `UPDATE` case in `dataProvider.js` builds the PATCH body from `deep-diff(previousData.$staged, data.$staged)`. The activation object in an IS-05 PATCH is a *request*, not staged state — but when a receiver's staged endpoint already reports the same activation values (e.g. a node that initializes staged `activation.mode` to `"activate_immediate"`, as in the reporter's setup), the diff contains no activation entries, so the Connect tab's Activate button and Staged/Edit immediate activation send a PATCH with **no** `activation` object, and nothing activates.

Minimal repro of the mechanism (deep-diff 1.0.2, same as the lockfile):

```js
const diff = require('deep-diff');
const prev = { activation: { mode: 'activate_immediate', requested_time: null }, master_enable: true };
const next = { activation: { mode: 'activate_immediate', requested_time: null }, master_enable: true, sender_id: 'abc' };
diff(prev, next); // => [{ kind: 'N', path: ['sender_id'], rhs: 'abc' }] — activation gone
```

## Fix

After building `patchData` from the differences, include `data.$staged.activation` whenever its `mode` is non-null. A null mode is deliberately *not* force-included: it is still only sent when it results from an actual difference, so pending scheduled activations are not unintentionally cancelled by unrelated edits.

## Verification

- `yarn lint-check` clean
- `yarn build` succeeds (production build)
- the repro above plus manual tracing of both affected paths (`makeConnection.js` Connect-tab activation and Staged/Edit) — both now always carry `activation` in the PATCH body when a mode is chosen

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Development/src/dataProvider.js
+++ b/Development/src/dataProvider.js
@@ -568,10 +568,12 @@ const convertDataProviderRequestToHTTP = (
                 JsonPointer.set(patchData, `/${d.path.join('/')}`, d.rhs, true);
             }
 
-            // the activation object in a PATCH requests an activation, rather
-            // than reflecting staged state, so when a mode is set, include it
-            // even if the staged endpoint already reported the same values,
-            // which otherwise leaves it out of the differences (see #97)
+            // Activation in a PATCH is a request, not staged state. Include it whenever
+            // the request has a non-null mode, even if GET /staged already reported the
+            // same values (otherwise deep-diff omits it). That also covers non-conformant
+            // devices that leave activate_immediate on /staged (see #97). A null mode is
+            // only sent when it appears in the differences, so unrelated edits do not
+            // cancel a pending activation.
             if (get(params, 'data.$staged.activation.mode') != null) {
                 set(
                     patchData,
--- a/Development/src/dataProvider.js
+++ b/Development/src/dataProvider.js
@@ -568,6 +568,18 @@ const convertDataProviderRequestToHTTP = (
                 JsonPointer.set(patchData, `/${d.path.join('/')}`, d.rhs, true);
             }
 
+            // the activation object in a PATCH requests an activation, rather
+            // than reflecting staged state, so when a mode is set, include it
+            // even if the staged endpoint already reported the same values,
+            // which otherwise leaves it out of the differences (see #97)
+            if (get(params, 'data.$staged.activation.mode') != null) {
+                set(
+                    patchData,
+                    'activation',
+                    get(params, 'data.$staged.activation')
+                );
+            }
+
             if (has(patchData, 'transport_file')) {
                 if (get(patchData, 'transport_file.data') === null) {
                     set(patchData, 'transport_file.type', null);


garethsb:
As pointed out in #97, the bug is in the Receiver in that case, not nmos-js, so needs further discussion about whether to include this workaround in nmos-js.

Development/src/dataProvider.js
```diff
@@ -568,6 +568,18 @@ const convertDataProviderRequestToHTTP = (
                 JsonPointer.set(patchData, `/${d.path.join('/')}`, d.rhs, true);
             }
 
+            // the activation object in a PATCH requests an activation, rather
+            // than reflecting staged state, so when a mode is set, include it
+            // even if the staged endpoint already reported the same values,
+            // which otherwise leaves it out of the differences (see #97)
```
garethsb:
```suggestion
            // Activation in a PATCH is a request, not staged state. Include it whenever
            // the request has a non-null mode, even if GET /staged already reported the
            // same values (otherwise deep-diff omits it). That also covers non-conformant
            // devices that leave activate_immediate on /staged (see #97). A null mode is
            // only sent when it appears in the differences, so unrelated edits do not
            // cancel a pending activation.
```

garethsb (approved):
I think by expanding the comment slightly, the rationale is a bit clearer. Then, seems OK to me. Thanks for the contribution.

ZayanKhan-12:
Thanks for the review @garethsb — I've applied your suggested comment wording verbatim in 6a391cc. I also ran the full controller test suite locally (IS-04-04, IS-05-03 and BCP-007-03-02 via the Testing Façade against the AMWA testing tool) and all three suites pass with this change.

lo-simon (approved):
LGTM

jonathan-r-thorpe:
@ZayanKhan-12 glad to hear you're using the Controller test suites with the Testing Façade 👍

PR #167, opened 2026-07-28
Update GitHub Actions to versions which run on Node.js 24

mattlamb99:
Recent runs of build-test emit warnings because the pinned actions target the deprecated Node.js 20 runtime and are being forced to run on Node.js 24, see https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/.

Update the pinned actions to their current major versions, which all target Node.js 24:

- actions/checkout v2/v3 to v7
- actions/setup-node v4 to v7
- actions/setup-python v4 to v7
- actions/upload-artifact v4 to v7
- actions/download-artifact v4 to v8

These are drop-in upgrades for this workflow. The artifact actions have shared the same backend since v4, so make_badges still downloads the artifacts uploaded by build_and_test, and setup-python still provides Python 3.8 on ubuntu-22.04.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -13,9 +13,9 @@ jobs:
   build_and_test:
     runs-on: ubuntu-22.04
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v7
 
-    - uses: actions/setup-node@v4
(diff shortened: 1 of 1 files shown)


lo-simon (approved):
LGTM

PR #169, opened 2026-07-30
Correcting #163 to exclude `activation_time` from /staged PATCH

lo-simon:
Correcting #163 

`activation_time` should not be sent in /staged PATCH requests for IS-05.

In IS-05, the activation object fields are:

- mode
- requested_time

So this new logic should include activation but sanitise it to only [schema-valid fields](https://specs.amwa.tv/is-05/releases/v1.1.2/APIs/schemas/resolved/activation-schema.html) (at minimum, exclude activation_time).

### DIFF
--- a/Development/src/dataProvider.js
+++ b/Development/src/dataProvider.js
@@ -8,7 +8,7 @@ import {
     UPDATE,
     fetchUtils,
 } from 'react-admin';
-import { assign, get, has, isEmpty, set } from 'lodash';
+import { assign, get, has, isEmpty, pick, set } from 'lodash';
 import { JsonPointer } from 'json-ptr';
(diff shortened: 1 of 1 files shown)


garethsb (approved):
Sounds correct and looks plausible. @jonathan-r-thorpe should/could this have been caught by controller tests somehow?

jonathan-r-thorpe:
@garethsb yes it could and should. I'll look at adding schema validation into the mock Nodes on the NMOS Testing Controller tests to trap this class of error.

jonathan-r-thorpe:
This error should be trapped with this addition to NMOS Testing https://github.com/AMWA-TV/nmos-testing/pull/904

PR #171, opened 2026-08-05
Fix Connection API Bridge routing, HEAD, and Location redirects

garethsb:
## Summary

Fixes Connection API Bridge routing and browser-facing behaviour when proxying Device Connection APIs through Envoy:

- Match bridge routes with `path_separated_prefix` so the client path after the Connection API version (none, `/`, or a sub-path) is preserved when rewriting onto the Device `basePath`. Trailing-slash handling stays with the upstream Device per IS-04/IS-05.
- Allow `HEAD` on bridge routes like `GET` (including retries) and advertise `HEAD` in the CORS allow-methods list.
- Rewrite upstream 3xx `Location` headers so browsers stay on the bridge for Connection API redirects that remain under that target's Connection API base path. Context (base path, bridge path, upstream scheme, candidate authorities) is injected per route via `LuaPerRoute` into dynamic metadata for `location_rewrite.lua` (Envoy Lua cannot read the LB-selected upstream host; route `filter_metadata` / `request_headers_to_add` were not usable for this on Envoy 1.31).

### Location policy

- **Path-relative / root-relative** resolved against the reconstructed upstream Connection API path: rewrite onto the bridge when the result stays under `base_path`; otherwise **502** with NMOS error JSON and `x-nmos-bridge-error` (cannot re-absolutize without knowing which candidate Envoy selected).
- **Absolute** or **scheme-relative** (filled with the client scheme): if scheme+authority match a candidate and the path is under `base_path`, rewrite onto the bridge; otherwise leave unchanged (including candidate URLs outside the base path, e.g. `/x-manifest/` or a `/transportfile` target on another path).
- Envoy **internal redirects** are not used: a fully qualified Device `Location` under `/x-nmos/` would be re-matched by path onto the Registry `/x-nmos/` routes.

Upstream scheme for matching comes from the Device control `href` (Phase 1 still only allows `http:`).

## Test plan

- [x] CI / `yarn lint-check` green
- [x] Bridge: `.../connection/v1.x`, `.../v1.x/`, and `.../v1.x/single/...` all reach the Device correctly
- [x] `HEAD` on a bridge Connection API path succeeds (and is allowed by CORS)
- [x] Candidate absolute / scheme-relative `Location` **under** Connection `base_path` is rewritten onto the bridge; foreign authorities left unchanged
- [x] Path-/root-relative under Connection base → bridge path; out-of-base relatives → 502 NMOS error; candidate absolute outside base (e.g. `/x-manifest/`) left unchanged

### DIFF
--- a/ConnectionBridge/README.md
+++ b/ConnectionBridge/README.md
@@ -164,4 +164,16 @@ The nmos-js client offers a **Connection Bridge** setting with three modes:
 
 ## Status
 
-Phase 1 (HTTP browser and upstream access, file-based dynamic configuration, `GET`/`POST`/`PATCH`, no response rewriting) is implemented, plus health checking and multi-endpoint failover from Phase 2. Not yet implemented: response size limits, HTTPS upstreams, authentication translation, mTLS, and an xDS control plane.
+Phase 1 is implemented, plus health checking and multi-endpoint failover from Phase 2:
+
+- HTTP browser and upstream access, file-based dynamic configuration
+- `GET`/`HEAD`/`POST`/`PATCH`
+- Upstream 3xx `Location` handling (see below)
+
+Not yet implemented: response size limits, HTTPS upstreams, authentication translation, mTLS, and an xDS control plane.
+
+`Location` handling uses each target's Connection API `base_path` (the path of the Device control `href`, typically `/x-nmos/connection/v1.1` or similar):
+
+- Absolute or scheme-relative (filled with the client scheme) Locations whose scheme and authority match a Device Connection API candidate and whose path stays under that `base_path` are rewritten onto the bridge; other absolute Locations are forwarded unchanged (including candidate URLs outside `base_path`, e.g. `http://device/x-manifest/...`).
+- Path-relative and root-relative Locations are resolved against the upstream Connection API path and rewritten onto the bridge when they stay under `base_path`; relatives outside `base_path` are rejected with `502` and an NMOS error body (`x-nmos-bridge-error` describes the unsupported Location), since an absolute Device URL cannot be reconstructed without knowing which candidate Envoy selected.
+- Envoy internal redirects are not used: absolute Device Locations under `/x-nmos/` would be re-routed by path onto the Registry `/x-nmos/` routes.
--- a/ConnectionBridge/README.md
+++ b/ConnectionBridge/README.md
@@ -83,6 +83,7 @@ Location rewrite policy is covered by Lua unit tests (Lua 5.1 / LuaJIT, matching
 
 ```bash
 lua5.1 location_rewrite_test.lua
+lua5.1 location_rewrite_test.lua -v   # per-case status lines
 ```
 
 ## Running
--- a/ConnectionBridge/README.md
+++ b/ConnectionBridge/README.md
@@ -77,6 +77,14 @@ Envoy health checks the candidates and fails over from higher to lower priority 
 
 Controls that are not safe to proxy are logged and ignored: missing or malformed hrefs, unsupported schemes (Phase 1 supports `http` upstreams only), hrefs whose path is inconsistent with the advertised version, and duplicates after normalization.
 
+## Testing
+
+Location rewrite policy is covered by Lua unit tests (Lua 5.1 / LuaJIT, matching Envoy). From `envoy/`:
+
+```bash
+lua5.1 location_rewrite_test.lua
+```
+
 ## Running
 
 ```bash
--- a/ConnectionBridge/README.md
+++ b/ConnectionBridge/README.md
@@ -37,20 +37,33 @@ Methods are restricted to `GET`, `POST`, `PATCH` and `OPTIONS`. Query strings, m
 ```text
 Browser
     |
-    v
- Envoy
+    +--(HTTP / WebSocket)------> Registry Query API
     |
-    +--> Registry APIs
+    +--(HTTP)------------------> Envoy
+                                    |
+                                    +--> /x-nmos-bridge/... --> Device Connection APIs
+                                    |
+                                    +--> /x-nmos -> ["query/"] (fixed listing)
+                                    |
+                                    +--> /x-nmos/query/... (convenience)
+                                    |         --> Registry Query API
+                                    |
+                                    +--> /x-dns-sd/... (convenience)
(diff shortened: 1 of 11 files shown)


garethsb:
👀 @mattlamb99

garethsb:
Hmm, I'm not sure the approach to handling redirects is going to work for all upstream APIs. The only explicit cases in the NMOS specs are for the `/transportfile` endpoint and for handling of paths with and without trailing slashes. But the problem is that the Lua downstream filter doesn't have access to the upstream request URL so it can't resolve non-absolute `Location` headers with any confidence and just returning an absolute or relative path or a scheme-relative URL is wrong. It may be better to turn on Envoy internal redirects(which currently only handles absolute locations itself...) and turn others into `502` Bad Gateway... Hmm.

garethsb:
Redirect handling now redone. I think I'm happy this will work for enough cases to be useful without returning clearly wrong redirect responses downstream.

ConnectionBridge/README.md
```diff
@@ -164,4 +164,4 @@ The nmos-js client offers a **Connection Bridge** setting with three modes:
...
...le-based dynamic configuration, `GET`/`HEAD`/`POST`/`PATCH`, `Location` handling on upstream 3xx) is implemented, plus health checking and multi-endpoint failover from Phase 2. Not yet implemented: response size limits, HTTPS upstreams, authentication translation, mTLS, and an xDS control plane. Absolute `Location` headers that match a Device Connection API candidate and stay under that Connection API base path are rewritten onto the bridge; other absolute Locations (including candidate URLs outside that base path, e.g. `/x-manifest/`) are forwarded unchanged. Scheme-relative Locations are resolved with the client scheme and then treated the same as absolute. Path-relative and root-relative Locations are resolved against the upstream Connection API path and rewritten onto the bridge when they stay under that base path; relatives outside that base path are rejected with `502` and an NMOS error body (`x-nmos-bridge-error` describes the unsupported Location), since an absolute Device URL cannot be reconstructed without knowing which candidate Envoy selected. Envoy internal redirects are not used: absolute Device Locations would be re-routed by path onto the Registry `/x-nmos/` routes.
```
jonathan-r-thorpe: > other absolute Locations (including candidate URLs outside that base path, e.g. `/x-manifest/`)

`/x-manifest/` is not an absolute path.

ConnectionBridge/envoy/location_rewrite.lua
```diff
@@ -0,0 +1,405 @@
...
+    end
+
+    -- root-relative (path-absolute)
+    if location:sub(1, 1) == "/" and location:sub(2, 2) ~= "/" then
+        local rewritten = rewrite_onto_bridge(location, base_path, bridge_path)
+        if rewritten ~= nil then
+            return rewritten
+        end
+        return "reject"
+    end
```
jonathan-r-thorpe, line 302: Am I right in thinking that a redirect to `/x-manifest/` will be rejected, but a redirect to `http://device/x-manifest/` will fall through?

ConnectionBridge/envoy/location_rewrite.lua
jonathan-r-thorpe, line 1: How difficult would it be to introduce some unit test for these Lua policies? There's a lot of URL processing that would be good to prove and catch future regressions.

ConnectionBridge/README.md
garethsb: What I meant was that if 

`http://device/x-nmos/connection/v1.2/single/sender/<id>/transportfile`
redirected to, for example,
`Location:  http://device/x-manifest/<id>` (an absolute URL)

then we should not turn that into a redirect onto the bridge, because `/x-manifest` is outside the base path `/x-nmos/connection/v1.2` that we're proxying.

ConnectionBridge/envoy/location_rewrite.lua
garethsb: Exactly

ConnectionBridge/envoy/location_rewrite.lua
garethsb: I have a whole bunch of test cases, but wiring them into CI seemed a bit painful... I'll have one more look.

ConnectionBridge/README.md
jonathan-r-thorpe:
```suggestion
Phase 1 (HTTP browser and upstream access, file-based dynamic configuration, `GET`/`HEAD`/`POST`/`PATCH`, `Location` handling on upstream 3xx) is implemented, plus health checking and multi-endpoint failover from Phase 2. Not yet implemented: response size limits, HTTPS upstreams, authentication translation, mTLS, and an xDS control plane. Absolute `Location` headers that match a Device Connection API candidate and stay under that Connection API base path are rewritten onto the bridge; other absolute Locations (including candidate URLs outside that base path, e.g. `http://device/x-manifest/`) are forwarded unchanged. Scheme-relative Locations are resolved with the client scheme and then treated the same as absolute. Path-relative and root-relative Locations are resolved against the upstream Connection API path and rewritten onto the bridge when they stay under that base path; relatives outside that base path are rejected with `502` and an NMOS error body (`x-nmos-bridge-error` describes the unsupported Location), since an absolute Device URL cannot be reconstructed without knowing which candidate Envoy selected. Envoy internal redirects are not used: absolute Device Locations would be re-routed by path onto the Registry `/x-nmos/` routes.
```

garethsb: Suggestion incorporated (and wall of text turned into bullet list). Thanks!

ConnectionBridge/envoy/location_rewrite.lua
garethsb: Not painful. Done. 😃

mattlamb99:
Nice, the reconstructed-upstream-path approach resolves the relative `Location` concern neatly. I tested the reject path and found one issue: Envoy's `body()` returns `nil` on a body-less 3xx (which redirects usually are), so replacing the response with the `502` NMOS error raises a Lua error instead. One-word fix (`body(true)`) plus unit tests for the `envoy_on_request`/`envoy_on_response` entry points via mock handles, targeted at your branch: garethsb/nmos-js#1

garethsb:
Thanks, @mattlamb99, merged and fixed `Content-Length` on top!

lo-simon (approved):
LGTM

PR #174, opened 2026-08-17
MXL Domain ID patch fix

jonathan-r-thorpe:
- When patching in a multi-node cluster environment where flows are replicated across nodes, the MXL domain a Sender writes to may be different to the MXL domain the Receiver reads from.
- For this reason it is unsafe to patch the domain id verbatim, and instead "auto" should be used
- This assumes that there is either one MXL domain per cluster node, or at most one MXL domain per media function meaning there is never ambiguity for Receiver's when selecting which MXL domain from which to read.


garethsb (approved):
LGTM

lo-simon (approved):
LGTM

PR #175, opened 2026-08-18
Bridge IS-08 Channel Mapping through the Connection API Bridge

garethsb:
## Summary
- Proxy Device `cm-ctrl` (IS-08 Channel Mapping) controls through Envoy alongside existing `sr-ctrl` (IS-05 Connection) targets, keyed by Device + API + version, with `DELETE` on the mutating-method allow-list.
- Make the whole `/x-nmos-bridge` namespace answer consistently (root / version listings and a namespace fall-through), and list the APIs and versions the bridge proxies for a given Device ID.
- Design plan: `ConnectionBridge/docs/channelmapping-proxy-plan.md` (Forced/Auto client remap still follow-up).

**Follow-up:** rename `ConnectionBridge` (directory, docs, and “Connection API Bridge” naming) to something like `nmos-bridge`, now that the front door is not Connection-only. Kept out of this PR to avoid churn on the behavioural change.

## Test plan
- [x] Adapter generates separate clusters/routes for IS-05 and IS-08 on the same Device (no target merge on shared v1.0)
- [x] `GET`/`POST`/`PATCH`/`DELETE` through `/x-nmos-bridge/v1.0/devices/{id}/channelmapping/v1.0/...` against an nmos-cpp (or other) node with `cm-ctrl`
- [x] Unknown bridge paths return NMOS-style JSON 404s (not the SPA / body-less Envoy 404)
- [x] `GET /x-nmos-bridge`, `/x-nmos-bridge/v1.0`, `/devices/{id}`, `/devices/{id}/{api}` return the expected listings
- [x] Existing IS-05 Connection bridge behaviour unchanged (inertness check on generated CDS/RDS excluding new channelmapping entries)
- [x] Lua Location rewrite unit tests still pass (including Channel Mapping cases)

### DIFF
--- a/ConnectionBridge/README.md
+++ b/ConnectionBridge/README.md
@@ -1,13 +1,13 @@
 # NMOS Connection API Bridge
 
-Provides browser-accessible proxy access to [AMWA IS-05](https://specs.amwa.tv/is-05/) Connection APIs exposed by Devices registered in an NMOS Registry, where the browser may not have network access to the Device APIs directly.
+Provides browser-accessible proxy access to [AMWA IS-05](https://specs.amwa.tv/is-05/) Connection APIs and [AMWA IS-08](https://specs.amwa.tv/is-08/) Channel Mapping APIs exposed by Devices registered in an NMOS Registry, where the browser may not have network access to the Device APIs directly.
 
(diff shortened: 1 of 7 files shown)


lo-simon (approved):
LGTM

Just one small comment. In the Architecture section of the README, /x-nmos-bridge/... can map to Device Connection APIs and Channel Mapping APIs.
```
Envoy
                                    |
                                    +--> /x-nmos-bridge/... --> Device Connection APIs and Channel Mapping APIs
```

PR #176, opened 2026-08-19
Rename ConnectionBridge to nmos-bridge

garethsb:
The bridge already proxies Channel Mapping (and Query/DNS-SD/Logging) as well as Connection, so the Connection-only directory and settings names were misleading. Keep the public `/x-nmos-bridge` path.


nmos-bridge/README.md
```diff
@@ -171,7 +171,7 @@ A layout that also proxies Query and Logging through Envoy might use:
 ```text
 Query API:              http://controller.example.com:8080/x-nmos/query/v1.3
 Logging API:            http://controller.example.com:8080/log/v1.0
-Connection Bridge API:  http://controller.example.com:8080/x-nmos-bridge/v1.0
+NMOS Bridge API:  http://controller.example.com:8080/x-nmos-bridge/v1.0
```
lo-simon: correct ident

```suggestion
NMOS Bridge API:   http://controller.example.com:8080/x-nmos-bridge/v1.0
```

lo-simon (changes requested):
LGTM

lo-simon (approved):

PR #178, opened 2026-08-20
Make IS-08 Channel Mapping editable

garethsb:
## Summary
- Add Device Edit for Channel Mapping: local draft on Active Map, Activate posts only the changed channels through `dataProvider` `UPDATE` of `devices`.
- Soft-validate `routable_inputs`, `block_size` and `reordering` on the matrix without blocking Activate, so a Node can still be tested with a predicted-invalid POST.
- Add scheduled activation modes (`requested_time`) matching Connection Edit, a pending Activations show tab, and Cancel via `DELETE …/map/activations/{id}`.

## Test plan
- [x] Immediate Activate against `nmos-cpp-node` updates Active Map and a later Edit seeds from that map.
- [x] Mapping cells that violate `routable_inputs` / incomplete blocks / non-fixed offset show the highest-priority warning; Activate still posts.
- [x] Relative schedule with a long `requested_time` (e.g. `300:0`) appears on Activations and can be cancelled.
- [x] No Bridge / Auto Bridge / Forced Bridge still use the resolved `$channelmappingAPI`.
- [x] `yarn test` in `Development/` (Jest now fails if no tests are found).

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -48,7 +48,7 @@ jobs:
       run: |
         yarn run lint-check
         yarn run build
-        yarn test --passWithNoTests --coverage --watchAll=false
+        yarn test --coverage --watchAll=false
 
     - name: install python
       uses: actions/setup-python@v7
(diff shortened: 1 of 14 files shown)


lo-simon (approved):
LGTM

PR #183, opened 2026-09-18
Add inline IS-13 Annotation API editing on resource Show pages

garethsb:
## Summary
- Discover each Node's Annotation API from its services and PATCH `label`, `description`, and `tags` from the Query record in place on Show pages (nodes, devices, sources, flows, senders, receivers).
- Pause auto-refresh while those fields (and Channel Mapping infotip custom names) are being edited, matching how dedicated Edit pages already behave.
- Query version gates for `description`/`tags` stay on the Node and Device Show pages; the Annotation container only handles discovery and write-back.

image

## Test plan
- [x] Open a Node that advertises IS-13; edit label/description (Enter to publish, Escape to cancel) and tags (add/edit/reset one, Reset all).
- [x] Confirm a rejected read-only tag stays in the editor and shows the API error, and that Query auto-refresh pauses while editing.
- [x] Confirm resources whose Node has no Annotation API still show the same fields read-only.
- [x] Confirm Channel Mapping header infotip name editing also pauses auto-refresh.


lo-simon (approved):
LGTM

